PR #5, opened 2015-04-30
Support random uid

csrwng:
Initial attempt at supporting non-root user

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -27,8 +27,6 @@ RUN yum install -y \
 
 ENV PATH $PATH:/opt/rh/php55/root/usr/bin
 
-USER root
-
 EXPOSE 8080
 
 # Copy the STI scripts from the specific language image to /usr/local/sti
@@ -38,6 +36,12 @@ COPY ./.sti/bin/ /usr/local/sti
 # run and build the applications.
 COPY ./contrib/ /opt/openshift
 RUN chown -R default:default /opt/openshift /tmp/sessions
+
+RUN sed -i -f /opt/openshift/httpdconf.sed /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf  && \
+    chmod -R og+rwx /opt/rh/php55/root/etc && \
+    chmod -R og+rwx /opt/rh/httpd24/root/var/run/httpd
+
+USER default
 
 # Set the default CMD to print the usage of the language image
 CMD ["usage"]
--- a/5.5/contrib/httpdconf.sed
+++ b/5.5/contrib/httpdconf.sed
@@ -1,0 +1,7 @@
+s/^Listen 80/Listen 0.0.0.0:8080/
+s/^User apache/User default/
+s/^Group apache/Group default/
+s%^DocumentRoot "/opt/rh/httpd24/root/var/www/html"%DocumentRoot "/opt/openshift/src"%
+s%^<Directory "/opt/rh/httpd24/root/var/html"%<Directory "/opt/openshift/src"%
+s%^ErrorLog "logs/error_log"%ErrorLog "/tmp/error_log"%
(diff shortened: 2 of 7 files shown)


csrwng:
@bparees ptal ... it's my naive first shot at running as non-root

5.5/Dockerfile
```diff
@@ -25,6 +25,12 @@ RUN yum install -y \
     yum clean all -y && \
     mkdir /tmp/sessions
 
+RUN chmod -R og+rwx /opt/rh/php55/root/etc && \
+    chmod -R og+rwx /opt/rh/httpd24/root/var/run/httpd && \
+    chmod -R og+rwx /opt/rh/httpd24/root/etc/httpd/logs && \
+    cat /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf | sed "s/^Listen 80/#Listen 80/" > /tmp/httpd.conf && \
```
bparees: sed -i

and since you're commenting out listen on 80, what's it listening on?

bparees:
does this work as is?  @jhadvig indicated there were problems writing the PID file, but maybe it's writing it to one of the paths you chmod'd.

5.5/Dockerfile
csrwng: 8080 ... it's in the httpd.conf we copy over

csrwng:
it does work as is... however not ideal... because the sti build still happens as root. However, the resulting image can be run as any user

csrwng:
@bparees - ready for a more serious review. Made the change to rhel image and updated tests to try with a random uid. Requires https://github.com/openshift/source-to-image/pull/177

5.5/Dockerfile
```diff
@@ -25,9 +25,18 @@ RUN yum install -y \
     yum clean all -y && \
     mkdir /tmp/sessions
 
-ENV PATH $PATH:/opt/rh/php55/root/usr/bin
+RUN sed -i -e 's/^Listen 80/Listen 0.0.0.0:8080/' \
```
bparees: I wonder if it's worth putting this in an external script that we ADD and RUN.  what do you think?  (you can use the contrib dir I believe)

csrwng:
@bparees after looking at the sti-perl repo, I think a better approach is to simply comment out what we don't need from the base httpd.conf and merge the one in the contrib directory. That way I can be more consistent with both projects.

5.5/Dockerfile
csrwng: yup, makes sense

csrwng:
updated to use an external .sed file for common httpd.conf replacements

csrwng:
[test]

5.5/contrib/httpdconf.sed
```diff
@@ -0,0 +1,7 @@
+s/^Listen 80/Listen 0.0.0.0:8080/
+s/^User apache/User default/
+s/^Group apache/Group default/
+s%^DocumentRoot "/opt/rh/httpd24/root/var/www/html"%DocumentRoot "/opt/openshift/src"%
+s%^<Directory "/opt/rh/httpd24/root/var/html"%<Directory "/opt/openshift/src"%
+s%^ErrorLog "logs/error_log"%ErrorLog "/tmp/error_log"%
+s%CustomLog "logs/access_log"%CustomLog "/tmp/access_log"%
```
bparees: can we bounce these logs to STDOUT instead?

bparees: actually i guess let's not, for now..but we probably have to decide what to do about it at some point.

bparees:
lgtm but would like to sort out why it seems to be failing the sti-php test build (possibly just the bad STI version issue you mentioned)

bparees:
rerunning the test here:
https://ci.openshift.redhat.com/jenkins/view/Image%20Verification/job/sti-php/10/

hopefully that picks up the new AMI that should have your STI changes.

PR #13, opened 2015-05-19
use uid instead of username

bparees:


bparees:
@csrwng ptal.

csrwng:
LGTM

hhorak:
@bparees What was the reason for changing this? Can you share, please?

bparees:
@hhorak we're going to be validating what user an image runs with so we can prevent images that try to run as root.  the only way we can sanely do this is to ensure the USER is not "0" since in theory any username could be mapped to uid 0 and we won't easily be able to look up the username in /etc/passwd to determine if they are a superuser.

So to that end, we're moving to specifying uids instead.  i've also documented this best practice here:
https://github.com/openshift/openshift-docs/pull/382/files

PR #18, opened 2015-06-03
Remove misspelled mcrypt

jhadvig:
Since the mcrypt pkg is not part of php55 scl we can remove it.
@bparees PTAL


bparees:
lgtm.

PR #19, opened 2015-06-04
Adding support for .htaccess file in app source

jhadvig:
@mfojtik @rhcarvalho PTAL


5.5/Dockerfile
```diff
@@ -35,6 +35,7 @@ RUN yum install -y \
     mkdir /tmp/sessions && \
     chown -R default:default /opt/openshift /tmp/sessions && \
     sed -i -f /opt/openshift/httpdconf.sed /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf && \
+    head -n151 /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf | tail -n1 | grep "AllowOverride All" || exit && \
```
mfojtik: this is pretty ugly but I don't see any other way to do this...

PR #20, opened 2015-06-10
fix typo in search string

bparees:


bparees:
@mfojtik this was the other pusher job issue.  same problem @ewolinetz hunted down in the perl image, except this time we didn't have the right config in place because of a bad search string in our sed script. :(

bparees:
and yes both are needed...centos uses one, rhel uses the other...  [sigh]

PR #25, opened 2015-06-11
remove pecl as it is not signed

bparees:


bparees:
@brenton @mfojtik will need to look into this later but i'm removing it so we can successfully build again for now.

bparees:
@brenton looks like the issue is that since you added the pecl rpm to li, we're picking it up from there but the package is not signed.  Is there something you need to to do to get it signed, or should i be editing our dockerfile to trust it unsigned?

bparees:
(or is there a cert I need to add to our dockerfile?)

brenton:
I'm working to get them signed this morning.  Sorry about that!

PR #26, opened 2015-06-12
Using RHEL rpms

brenton:


brenton:
@mfojtik, @bparees PTAL

brenton:
The test failure seems to be AWS related.  Trying to kick off another: [test]

bparees:
thanks @brenton

PR #35, opened 2015-06-30
Restore old labels for RHEL7

mfojtik:


mfojtik:
@bparees PTAL

bparees:
why no centos compat labels?

PR #42, opened 2015-07-28
Add support for PHP 5.6

remicollet:
Notice, I drop php-fpm (not need, apache use mod_php), php-devel and imagick (not available in RHSCL)


bparees:
can you update the Makefile so it'll automatically test both versions?
https://github.com/openshift/sti-php/blob/master/Makefile#L2-L14

should look like:
https://github.com/openshift/mysql/blob/master/Makefile#L2-L16

hhorak:
I've taken a look and haven't spotted any issue, except one thing -- we should use `rhscl/rh-php56-rhel7` Name label instead of `openshift3/php-56-rhel7` for RHEL image. Bad is that after this change the usage message won't be accurate.

remicollet:
Image built in OSBS, so eveythin "should" be ok.

bparees:
needs a squash

bparees:
test

bparees:
[test]

5.6/.sti/bin/assemble
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+echo "---> Installing application source ..."
+cp -Rf /tmp/src/. ./
+
+if [ -f composer.json ]; then
+  echo "Installing dependencies from composer.json... "
```
mfojtik: Please make this message consistent with the message above :-) I know that I sounds like I nit picking, but in the end it makes the build logs in OpenShift look much nicer.

5.6/Dockerfile.rhel7
```diff
@@ -0,0 +1,61 @@
...
+
+LABEL io.k8s.description="Platform for building and running PHP 5.6 applications" \
+      io.k8s.display-name="Apache 2.4 with PHP 5.6" \
+      io.openshift.expose-services="8080:http" \
+      io.openshift.tags="builder,php,php56"
+
+LABEL k8s.io/description="Platform for building and running PHP 5.6 applications" \
+      k8s.io/display-name="Apache 2.4 with PHP 5.6" \
+      openshift.io/expose-services="8080:http" \
+      openshift.io/tags="builder,php,php56"
```
mfojtik: @hhorak should this image also get the rh-php56 tag?

mfojtik:
[test] again

5.6/Dockerfile.rhel7
```diff
@@ -0,0 +1,61 @@
+FROM rhel7
```
bparees: needs to be FROM openshift/base-rhel7 
that's why the test failed.

5.6/Dockerfile.rhel7
hhorak: I'd say so. if it is just for searching, somebody may try to find this image by asking for SCL name.

mfojtik: @remicollet can you add `rh-php6` here and to the label set above?

5.6/.sti/bin/assemble
remicollet: Bikeshedding... especially as this file is only a copy/paste from the 5.5 one ;)

5.6/Dockerfile.rhel7
remicollet: Will do

5.6/Dockerfile.rhel7
remicollet: But fails in OSBS

5.6/.sti/bin/assemble
mfojtik: You can also fix 5.5 :-)

5.6/Dockerfile.rhel7
bparees: then we need the openshift/base-rhel7 in OSBS....  
@sdodson how are you dealing with this today?

bparees:
@mfojtik it won't pass tests until the FROM rhel7 issue is addressed.

5.6/Dockerfile.rhel7
sdodson: We have a diff between this github repo and dist-git. We should probably discuss how we can minimize the diffs or automate them out somehow. If we no longer build RHEL images in jenkins then I'd say switch it to openshift3/sti-base.

bparees: we're still going to build them in out of this repo because we do our automated testing via this repo, so that's how we valdiate the PR doesn't break the rhel image.

we don't have a good way to do automated testing once we're in dist-git/OSBS.

5.6/Dockerfile.rhel7
```diff
@@ -0,0 +1,61 @@
+FROM openshift3/sti-base
```
bparees: this still needs to be:
FROM openshift/base-rhel7

remicollet: Sorry, but I'm totally lost in this base image nightmare.

openshift3/sti-base is used in other collections and build in OSBS

I have absolutely no idea of diff between openshift3/sti-base and openshift/base-rhel7. So I really need clear explanation, expectation, description (link to base image Dockerfile to understand), etc

In all case, I won't touch this before september (hollidays)

bparees: if that's the case we'll probably cherry pick this and deliver it in another PR.

To explain the base image "nightmare":  when we build these upstream repositories, we use a different set of docker registries than what OSBS is building and pushing to.

That means the sti-base upstream repo rhel image (openshift/base-rhel7), when built, gets pushed to an internal registry that we use for QE purposes and to ensure we can do a CI/CD flow.  The other images (eg this one) need to be based on that same openshift/base-rhel7 image so they can also be built+tested in that environment, continuously.

We then have the downstream dist-git repos which are completely internal and built by OSBS.  The output of OSBS builds are the openshift3/<whatever> images.  That means that when we pull code from this repo down into the dist-git repo, there are some changes that have to be made, including changing what the FROM image is, because the downstream(dist-git) repo for sti-base gets pushed to a different location (openshift3/sti-base?  i'm actually not 100% certain of the name offhand) when built by OSBS.

i hope that helps.  the main thing to keep in mind is that this repo and the dist-git repo do not have identical content.

bparees:
new pull is here: https://github.com/openshift/sti-php/pull/52

PR #43, opened 2015-07-31
fix opcache library path

bparees:


bparees:
@mfojtik ptal.

for https://bugzilla.redhat.com/show_bug.cgi?id=1248952

note that i could not reproduce the issue, but hopefully this fixes it.

bparees:
(I did end up recreating it, this appears to fix it)

PR #50, opened 2015-08-12
Update Dockerfile.rhel7

derdanu:
Default Session Path will be overwritten in this file.


5.5/Dockerfile.rhel7
```diff
@@ -53,6 +53,8 @@ RUN sed -i -f /opt/app-root/etc/httpdconf.sed /opt/rh/httpd24/root/etc/httpd/con
     chmod -R og+rwx /opt/rh/httpd24/root/var/run/httpd && \
     chown -R 1001:1001 /opt/app-root /tmp/sessions
 
+RUN sed -i '/php_value session.save_path/d' /opt/rh/httpd24/root/etc/httpd/conf.d/php55-php.conf
```
bparees: Thanks for catching this!

can you && this onto the previous RUN command rather than adding an additional layer?

also feel free to merge your other PR with this one rather than having two PRs.

bparees:
[test]

derdanu:
Merged in PR #53

PR #53, opened 2015-08-17
Fix: Default Session Path will be overwritten

derdanu:


bparees:
[test]

bparees:
i think this is failing because softwarecollections is currently down.  will try it again later.

bparees:
[test]

bparees:
[test]

bparees:
sorry for the delay, thanks for the fix!

PR #58, opened 2015-09-17
Update README with hot deploy

jhadvig:
Since php picks up changes in source by default, there is nothing else just update README && official docs.

@mfojtik @bparees thoughts ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,3 +211,16 @@ In case the **DocumentRoot** of the application is nested within the source dire
 users can provide their own **.htaccess** file.  This allows the overriding of Apache's behavior and
 specifies how application requests should be handled. The **.htaccess** file needs to be located at the root
 of the application source.
+
+Hot deploy
+---------------------
+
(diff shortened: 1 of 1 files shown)


mfojtik:
Please say where the sources are located inside the container, that will be the first question you will get on IRC when somebody will try to use hot-deploy ;-)

jhadvig:
@mfojtik updated :mountain_cableway:

bparees:
[test]

README.md
```diff
@@ -211,3 +211,16 @@ In case the **DocumentRoot** of the application is nested within the source dire
 users can provide their own **.htaccess** file.  This allows the overriding of Apache's behavior and
 specifies how application requests should be handled. The **.htaccess** file needs to be located at the root
 of the application source.
+
+Hot deploy
+---------------------
+
+Hot deploy works in the built image out of the box.
```
bparees: "Hot deploy works in this image out of the box."

PR #60, opened 2015-10-01
move newer image into centos/rhscl namespace

praiskup:
While we are on it, also unify the build scripts among images'
repositories:

- hack/ directory should be _just_ mirrored
- Makefile should be as declarative as possible
- build.sh should not rely on repository name


jhadvig:
[test]

PR #63, opened 2015-10-08
fix opcache config path, fix #62

### LINKED ISSUE #62: Wrong path to opcache.so module
Docker image for CentOS7 and php5.6 produce the following erorr to the `/tmp/error_log`:

```
Failed loading /opt/rh/php54/root/usr/lib64/php/modules/opcache.so:  /opt/rh/php54/root/usr/lib64/php/modules/opcache.so: cannot open shared object file: No such file or directory
```

And indeed:

```
$ docker exec 895ce6156408 ls -l /opt/rh/php54/root/usr/lib64/php/modules/opcache.so
ls: cannot access /opt/rh/php54/root/usr/lib64/php/modules/opcache.so: No such file or directory
```

Because it's in another directory:
```
$ docker exec 895ce6156408 ls -l /opt/rh/rh-php56/root/usr/lib64/php/modules/opcache.so
-rwxr-xr-x. 1 root root 154456 Jun 29 08:26 /opt/rh/rh-php56/root/usr/lib64/php/modules/opcache.so

```

I believe that it should be fixed in the following files:

```
$ git grep php54
5.6/contrib/etc/php.d/10-opcache.ini.template:zend_extension=/opt/rh/php54/root/usr/lib64/php/modules/opcache.so
5.6/contrib/etc/php.d/10-opcache.ini.template:opcache.blacklist_filename=/opt/rh/php54/root/etc/php.d/opcache*.blacklist
```

Steps to reproduce:
1) `$ make build VERSION=5.6 TARGET=centos7`
2) `s2i build https://github.com/openshift/sti-php.git --context-dir=5.6/test/test-app openshift/php-56-centos7 php-test-app`
3) `docker run -d -p 8080:8080 php-test-app`
4) `docker exec $(docker ps | grep php-test-app | awk '{print $1}') cat /tmp/error_log`

(Found during my work on BZ#1231862)

remicollet:
Since 5.5, "zend_extension" don't requires full path, so safer to fix both 5.5 and 5.6

### DIFF
--- a/5.5/contrib/etc/php.d/opcache.ini.template
+++ b/5.5/contrib/etc/php.d/opcache.ini.template
@@ -1,5 +1,5 @@
 ; Enable Zend OPcache extension module
-zend_extension=/opt/rh/php55/root/usr/lib64/php/modules/opcache.so
+zend_extension=opcache.so
 
 ; Determines if Zend OPCache is enabled
 opcache.enable=1
--- a/5.6/contrib/etc/php.d/10-opcache.ini.template
+++ b/5.6/contrib/etc/php.d/10-opcache.ini.template
@@ -1,5 +1,5 @@
 ; Enable Zend OPcache extension module
-zend_extension=/opt/rh/php54/root/usr/lib64/php/modules/opcache.so
+zend_extension=opcache.so
 
 ; Determines if Zend OPCache is enabled
 opcache.enable=1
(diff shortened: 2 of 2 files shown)


mfojtik:
LGTM, thanks @remicollet

jhadvig:
LGTM

PR #64, opened 2015-10-09
Configure Apache to write access logs to stdout instead of file

php-coder:
I've configured Apache to use [piped logs](http://httpd.apache.org/docs/2.4/logs.html#piped) and now it's writing access logs to stdout instead of `/tmp/access_log`. This change addresses to https://bugzilla.redhat.com/show_bug.cgi?id=1231862

PTAL @rhcarvalho @mfojtik
I've tested in on with php55/php56 and CentOS7.

A couple of related questions:
1) now there is no `/tmp/access_log`. Is it ok or we should provide backward compatibility? If we should I think we may change `cat` to `tee` and it would work fine
2) (if the answer to the first question is "No") Should we update our documentation like we did in https://github.com/openshift/openshift-docs/pull/887?

### DIFF
--- a/5.5/contrib/etc/httpdconf.sed
+++ b/5.5/contrib/etc/httpdconf.sed
@@ -5,5 +5,5 @@ s%^DocumentRoot "/opt/rh/httpd24/root/var/www/html"%DocumentRoot "/opt/app-root/
 s%^<Directory "/opt/rh/httpd24/root/var/www/html"%<Directory "/opt/app-root/src"%
 s%^<Directory "/opt/rh/httpd24/root/var/html"%<Directory "/opt/app-root/src"%
 s%^ErrorLog "logs/error_log"%ErrorLog "/tmp/error_log"%
-s%CustomLog "logs/access_log"%CustomLog "/tmp/access_log"%
+s%CustomLog "logs/access_log"%CustomLog "|/usr/bin/cat"%
 151s%AllowOverride None%AllowOverride All%
--- a/5.6/contrib/etc/httpdconf.sed
+++ b/5.6/contrib/etc/httpdconf.sed
@@ -5,5 +5,5 @@ s%^DocumentRoot "/opt/rh/httpd24/root/var/www/html"%DocumentRoot "/opt/app-root/
 s%^<Directory "/opt/rh/httpd24/root/var/www/html"%<Directory "/opt/app-root/src"%
 s%^<Directory "/opt/rh/httpd24/root/var/html"%<Directory "/opt/app-root/src"%
 s%^ErrorLog "logs/error_log"%ErrorLog "/tmp/error_log"%
-s%CustomLog "logs/access_log"%CustomLog "/tmp/access_log"%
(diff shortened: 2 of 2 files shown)


rhcarvalho:
I'll defer the answer to (1) to someone else, but answering (2), in either case we should update the docs to reflect the new reality.

mfojtik:
1) i don't think we have to keep backward compatibility for that file
2) we should say that all logs are streamed to container output (iow. stdout).

This change LGTM

mfojtik:
@remicollet @hhorak FYI

php-coder:
>2) we should say that all logs are streamed to container output (iow. stdout).

@mfojtik To be honest not all logs are streamed to container output. `/tmp/error_log` is still there, I didn't touch it. Should I replace using of `/tmp/error_log` by streaming to `/dev/stderr`?

mfojtik:
@php-coder we should stream all logs to stdout so they can be captured by some logging tools or directly by docker (`docker logs`). We can have option to log to volume in case you want to do some analysis in other container (?).

php-coder:
>(2), in either case we should update the docs to reflect the new reality

@rhcarvalho What did you mean when saying "we"? I'm just trying to understand what I should do on my side and what is not my side. Should I create a ticket for that? Or I should provide a pull request?

rhcarvalho:
@php-coder normally if you change the code / behavior, you're also responsible for updating the related documentation. Please provide a PR to update the docs.
I meant "we" in general, and in particular "you" :)

php-coder:
In private conversation with @mfojtik we came to the the following decision:
- current logging of access log to `/dev/stdout` is OK
- we don't want to stream error logs to `/dev/stderr` because we are not sure that is really needed and who/how will handle it
- we will add option to save access logs to the file by demand

@mfojtik also told me that there is a possible risk to expose sensitive information (like tokens in a URL). As far I understand we are not doing anything with it for a now.

mfojtik:
LGTM @php-coder can you check what we do for sti-perl ? (another httpd based image)

php-coder:
>Please provide a PR to update the docs.

Done: https://github.com/openshift/openshift-docs/pull/1058

php-coder:
>LGTM @php-coder can you check what we do for sti-perl ? (another httpd based image)

@mfojtik After a first look I can see that [sti-perl also writes logs to `/tmp/{access,error}_logs`](https://github.com/openshift/sti-perl/blob/9437495f4bcbdbee9d5966e598f0abc53f0f275d/5.20/contrib/etc/httpdconf.sed#L6). Would we fix it?

mfojtik:
@php-coder yes please (i'll merge this, but please open another PR for perl)

bparees:
> we will add option to save access logs to the file by demand

@mfojtik @php-coder doesn't seem like this ever happened, either here or in the perl repo that was also updated?

php-coder:
@bparees Perhaps I should rephrase myself: "we will add option to save access logs to the file if someone ask about it".

bparees:
@php-coder ok.  I kinda think it should have been there with this change (consider me the person asking about it) but we can live with it as is.

PR #67, opened 2015-10-22
One option to rule them all

remicollet:
Here is a proposal to allow more configuration of the image.

See https://github.com/openshift/sti-base/issues/69

The PHP_OPTIONS allows to set any PHP option
The HTTPD_OPTIONS allows to set any HTTPD option
The PHP_DISABLED_EXT allows ti disabled unneeded extension

Good test (low memory image) will be to use PHP_OPTIONS=memory_limit=32M, HTTPD_OPTIONS=MaxRequestWorkers=4 and PHP_DISABLED_EXT="gd intl ldap xdebug"

### DIFF
--- a/5.6/s2i/bin/run
+++ b/5.6/s2i/bin/run
@@ -20,4 +20,27 @@ export PHP_INI_SCAN_DIR=${PHP_INI_SCAN_DIR:-/etc/opt/rh/rh-php56/php.d}
 envsubst < /opt/app-root/etc/php.ini.template > $PHPRC
 envsubst < /opt/app-root/etc/php.d/10-opcache.ini.template > $PHP_INI_SCAN_DIR/10-opcache.ini
 
+HTTPD_CONFDIR=/opt/rh/httpd24/root/etc/httpd/conf.d
+
+if [ -n "$PHP_OPTIONS" ]; then
+  for opt in $PHP_OPTIONS
+  do    echo $opt
+  done >$PHP_INI_SCAN_DIR/99-options.ini
+fi
+
+if [ -n "$HTTPD_OPTIONS" ]; then
+  for opt in $HTTPD_OPTIONS
+  do    echo $opt
+  done >$HTTPD_CONFDIR/options.conf
+fi
+
+if [ -n "$PHP_DISABLED_EXT" ]; then
+  for $ext in $PHP_DISABLED_EXT
+  do
+    if [ -f $PHP_INI_SCAN_DIR/??-$ext.ini ]
+    then mv $PHP_INI_SCAN_DIR/??-$ext.ini $PHP_INI_SCAN_DIR/$ext.off
+    fi
+  done
+fi
+
 exec httpd -D FOREGROUND
--- a/5.6/s2i/bin/run
+++ b/5.6/s2i/bin/run
@@ -24,21 +24,22 @@ HTTPD_CONFDIR=/opt/rh/httpd24/root/etc/httpd/conf.d
 
 if [ -n "$PHP_OPTIONS" ]; then
   for opt in $PHP_OPTIONS
-  do    echo $opt
-  done >$PHP_INI_SCAN_DIR/99-options.ini
+  do php -r "echo urldecode('$opt') . PHP_EOL;"
+  done > $PHP_INI_SCAN_DIR/99-options.ini
 fi
 
 if [ -n "$HTTPD_OPTIONS" ]; then
   for opt in $HTTPD_OPTIONS
-  do    echo $opt
-  done >$HTTPD_CONFDIR/options.conf
+  do php -r "echo urldecode('$opt') . PHP_EOL;"
+  done > $HTTPD_CONFDIR/options.conf
 fi
 
 if [ -n "$PHP_DISABLED_EXT" ]; then
-  for $ext in $PHP_DISABLED_EXT
+  for ext in $PHP_DISABLED_EXT
   do
     if [ -f $PHP_INI_SCAN_DIR/??-$ext.ini ]
     then mv $PHP_INI_SCAN_DIR/??-$ext.ini $PHP_INI_SCAN_DIR/$ext.off
+    else echo "ERROR: $ext.ini not found in $PHP_INI_SCAN_DIR"
     fi
   done
 fi
--- a/5.6/s2i/bin/usage
+++ b/5.6/s2i/bin/usage
@@ -12,4 +12,26 @@ s2i build https://github.com/openshift/sti-php.git --context-dir=/5.6/test/test-
 
 You can then run the resulting image via:
 docker run -p 8080:8080 php-test-app
+
+Environment variables
+---------------------
+
+The image recognizes the following environment variables that you can set during
+initialization by passing `-e VAR=VALUE` to the Docker run command.
+
+|    Variable name           |    Description                                      |
+| -------------------------- | --------------------------------------------------- |
+| ERROR_REPORTING            | PHP error_reporting, default E_ALL & ~E_NOTICE      |
+| DISPLAY_ERRORS             | PHP display_error, default ON                       |
+| DISPLAY_STARTUP_ERRORS     | PHP display_startup_errors, default OFF             |
+| TRACK_ERRORS               | PHP track_errors, default OFF                       |
+| HTML_ERRORS                | PHP html_errors, default ON                         |
+| INCLUDE_PATH               | PHP include_path                                    |
+| SESSION_PATH               | PHP session_path, default /tmp/sessions             |
+| OPCACHE_MEMORY_CONSUMPTION | PHP opcache.memory_consumption, default 16M         |
+| OPCACHE_REVALIDATE_FREQ    | PHP opcache.revalidate_freq, default 2              |
+| PHP_OPTIONS                | PHP additional options, e.g. memory_limit=32M       |
+| PHP_DISABLED_EXT           | PHP disabled extensions, e.g. "xdebug intl"         |
+| HTTPD_OPTIONS              | HTTPD additional options, e.g. MaxRequestWorkers=16 |
+
 EOF
--- a/5.6/s2i/bin/usage
+++ b/5.6/s2i/bin/usage
@@ -19,19 +19,21 @@ Environment variables
 The image recognizes the following environment variables that you can set during
 initialization by passing `-e VAR=VALUE` to the Docker run command.
 
-|    Variable name           |    Description                                      |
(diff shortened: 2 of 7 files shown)


remicollet:
Notice, for the future, ideas:

- deprecates other options (ERROR_REPORTING, DISPLAY_ERRORS, ...) as  PHP_OPTIONS shoudl be enough
- disable uncommon ext and add the PHP_ENABLED_EXT option to enable them when needed

5.6/s2i/bin/usage
```diff
@@ -12,4 +12,26 @@ s2i build https://github.com/openshift/sti-php.git --context-dir=/5.6/test/test-
...
+| DISPLAY_STARTUP_ERRORS     | PHP display_startup_errors, default OFF             |
+| TRACK_ERRORS               | PHP track_errors, default OFF                       |
+| HTML_ERRORS                | PHP html_errors, default ON                         |
+| INCLUDE_PATH               | PHP include_path                                    |
+| SESSION_PATH               | PHP session_path, default /tmp/sessions             |
+| OPCACHE_MEMORY_CONSUMPTION | PHP opcache.memory_consumption, default 16M         |
+| OPCACHE_REVALIDATE_FREQ    | PHP opcache.revalidate_freq, default 2              |
+| PHP_OPTIONS                | PHP additional options, e.g. memory_limit=32M       |
+| PHP_DISABLED_EXT           | PHP disabled extensions, e.g. "xdebug intl"         |
+| HTTPD_OPTIONS              | HTTPD additional options, e.g. MaxRequestWorkers=16 |
```
bparees: the two _OPTIONS variables need better doc since they can take a list of options, right?  If i'm reading your bash code correctly, it's a space-separated list of a=b items.

mfojtik:
@remicollet can you make more examples when using multiple configuration settings for PHP_OPTIONS? How will you separate them? If space, how about escaping :-)

5.6/s2i/bin/run
```diff
@@ -20,4 +20,27 @@ export PHP_INI_SCAN_DIR=${PHP_INI_SCAN_DIR:-/etc/opt/rh/rh-php56/php.d}
 envsubst < /opt/app-root/etc/php.ini.template > $PHPRC
 envsubst < /opt/app-root/etc/php.d/10-opcache.ini.template > $PHP_INI_SCAN_DIR/10-opcache.ini
 
+HTTPD_CONFDIR=/opt/rh/httpd24/root/etc/httpd/conf.d
+
+if [ -n "$PHP_OPTIONS" ]; then
+  for opt in $PHP_OPTIONS
```
bparees: what does the syntax look like to specify multiple options in which some of the options i'm specifying have spaces in either the option name, or the option value?

remicollet: No option have space in name.
And I don't see any with space in value (usually boolean, interger or path).

5.6/s2i/bin/usage
remicollet: Please notice, than previous varibles have NO documentation ;)

And again, this proposal is for test and discussion. Please don't go to bikeshedding for now.

5.6/s2i/bin/run
bparees: i'd still like to see syntactical evidence that it could be supported with this approach.

bparees:
needs a test case.

5.6/s2i/bin/run
remicollet: i'd still like to see syntactical evidence that it could not be supported with this approach.

bparees: this is your PR and proposal, i'm not going to test your feature for you.  i'm not claiming it won't work, i'm just asking what it looks like so we understand what a user has to do if they need to define an env variable like that, and decide if that is an acceptable user experience.

remicollet:
Test, no options

    # docker run  -p 8080:8080  test56-app
    ps -ef | grep httpd => 5 process
    wget -q http://localhost:8080 -O - | grep xdebug.ini => found
    wget -q http://localhost:8080 -O - | grep memory_limit => 128M

Test, with options

    # docker run  -p 8080:8080 \
         -e PHP_OPTIONS=memory_limit=32M \
         -e PHP_DISABLED_EXT=xdebug \
         -e "HTTPD_OPTIONS=StartServers%202 MinSpareServers%202" test56-app
    ps -ef | grep httpd => 3 process (StartServers + master)
    wget -q http://localhost:8080 -O - | grep xdebug.ini => NOT found
    wget -q http://localhost:8080 -O - | grep memory_limit => 32M

bparees:
Can you add those tests to the test script?  that's what i meant by needs test cases.

ebalsumgo:
So how would I go about selecting my MPM with this? By default i need to use mpm_prefork in order to run PHP in non thread-safety mode. if i am to use the more efficient mpm_worker, php needs FPM and apache has to run fastcgi. Just as an example.

remicollet:
This change is not design to change MPM but only options.

Change MPM have no sense for PHP as mod_php used here is only for prefork.

Using FPM would a totally different way, and will propably require 2 containers, 1 with apache, 1 with FPM.

ebalsumgo:
@bparees i am still of the opinion that anything HTTPD_OPTIONS related should probably come from the _base rather than the images it build on it. 

@remicollet you are right, changing the php image to use mpm_worker would require a lot of work, puns!, but my general point was the availability of MPM selection and in turn any httpd related tech's performance.

bparees:
> @bparees i am still of the opinion that anything HTTPD_OPTIONS related should probably come from the _base rather than the images that build on it. 

@ebalsumgo we're not going to put it into the base image because that doesn't include httpd.  if/when we introduce a common httpd base image that is used for the frameworks that utilize httpd, that would make sense.

@remicollet the test script still needs to be updated to test this new behavior.

remicollet:
Sorry, but I have absolutely no idea why test is failing.

mfojtik:
@remicollet when you run them locally, will they pass?

remicollet:
> @remicollet when you run them locally, will they pass?
I'm unable to run test locally.
So help welcome to solve this.

bparees:
@remicollet you need to rebase your PR and that'll fix the broken test.

ebalsumgo:
@bparees you are right, i take back my statement, i thought the HTTPD was in base. Though I would say that moving all httpd related bits to a standard sti-httpd would be desired since as my original issue was saying, there are a bunch of scenarios where I would want to change MPM and tweak other things in order to suit the sti-ruby vs sti-php apps. So if there is anywhere I can "vote" or anything on creating such an image please share so I can be a nuisance there :)

@remicollet sorry for hijacking your PR

bparees:
@ebalsumgo https://trello.com/c/keOtxxXt/697-5-add-httpd-s2i-image-as-a-base-for-httpd-images

remicollet:
@bparees the test/run script is buggy ;)
CONTAINER_ARGS defined "after" begin used...

bparees:
> @bparees the test/run script is buggy ;)
CONTAINER_ARGS defined "after" begin used...

?

the usage is in a function and having it defined or not is optional.

remicollet:
> the usage is in a function and having it defined or not is optional.
But having it defined after the call of the function which use it is a bug ;)

bparees:
no it's not.......
```
$ cat /tmp/t.sh 
#!/bin/sh

foo() {
  echo $FOO
}

FOO=bar foo

sti-php (master)$ /tmp/t.sh 
bar
```

remicollet:
Yeah forget, I'm tired.

remicollet:
All seems ok now.

5.6/test/run
```diff
@@ -51,7 +51,8 @@ prepare() {
 }
 
 run_test_application() {
-  run_args=${CONTAINER_ARGS:-}
+  run_args="${CONTAINER_ARGS:-} -e PHP_OPTIONS=memory_limit=32M"
```
bparees: the whole point of the CONTAINER_ARGS env variable is that you set it when you invoke run_test_application, not override it within run_test_application.

5.6/test/run
```diff
@@ -162,10 +175,23 @@ test_application() {
 
   test_connection
   check_result $?
+
+  info "Checking the PHP_OPTIONS"
```
bparees: tests for HTTPD_OPTIONS and PHP_DISABLED_EXT?

bparees: you could docker run the image and then use docker cp to copy the config out of the container and inspect it.

bparees:
some more comments around testing and you can start to squash your commits.

remicollet:
> tests for HTTPD_OPTIONS 

I don't have any idea how to test this.

> PHP_DISABLED_EXT?

Bad idea, better to load the image with "all" extensions to ensure all work with default ext set.

I think, I have do my best.
Don't plan to spend more time on this.

bparees:
> I don't have any idea how to test this.

Like i said in my comment, you can run the image and then use docker cp to ensure the conf files got properly set up as you would expect them to be based on the options provided in the environment variable.

> Bad idea, better to load the image with "all" extensions to ensure all work with default ext set.

i don't know what you mean.  Sure by default you want it all enabled, but you've defined an environment variable that's supposed to control how the image runs, and you haven't provided a test case to confirm that the environment variable actually does what it's supposed to do (not only to prove it works now, but to ensure someone else doesn't break it later without realizing it).  So there needs to be a test case that runs the image using this environment variable and confirms that it does what is expected.


> I think, I have do my best.
Don't plan to spend more time on this.

i dont want to be a pain, but we aren't going to merge new features without unit tests that confirm they don't get broken later.  that's the only way we can move as quickly as we do:  by having automated tests to give us the confidence to quickly modify and re-spin images without worrying we've broken things, or at least to quickly catch it when we do.

@hhorak is there someone on your team who wants to finish up this PR?

hhorak:
@remicollet I agree with @bparees here that we should have at least simple coverage around features like this. Is there a reason why the `docker cp` test wouldn't work?  It seems to me it will be quite straightforward to do it.

remicollet:
Sorry, I spent too much time on this, and I'm discouraged but the lack of possible feedback.
As nobody seems interested. closing.

bparees:
@remicollet i don't think that's a fair characterization of what happened here.  The proposal certainly has merit, and you got a bunch of feedback about what is needed to complete the pull (specifically: tests for the new functionality, questions about how it would handle certain scenarios).

sebas5384:
@remicollet I just wanted to leave here that I needed exactly this PR.
Is 2017, there's anything like this PR at the current state of the php images ? I really don't want to create a specific Docker file just to override the php.ini file.

PR #81, opened 2016-01-12
Verify all installed packages using rpm -V

mfojtik:


mfojtik:
[test]

mfojtik:
@bparees more here:

```console
package php55-php-pecl-imagick is not installed
package php55-php-pecl-xdebug is not installed
```

@remicollet FYI

remicollet:
Those package are not part of RHSCL, so cannot be installed in the php 5.5 imahe.

mfojtik:
@remicollet ok, thanks, I will remove them then.

mfojtik:
@sdodson @bparees just out of curiosity, do we have those packages in openshift channels? I do remember talking about php something a long time ago. I think we should remove them from the centos image as they don't exists in centos scl, but I wonder if they exists in our channels.

mfojtik:
@remicollet how about php-5.6

remicollet:
5.6 (in RHSCL) only have Xdebug, no imagick.

PR #84, opened 2016-02-18
Pass all env to apache

mfojtik:
Fixes: https://github.com/openshift/sti-php/issues/83

@bparees @remicollet is this sane? (we pass all env in other images, so php should not be special here). This will allow users to do $_ENV

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -36,7 +36,10 @@ COPY ./contrib/ /opt/app-root
 RUN sed -i -f /opt/app-root/etc/httpdconf.sed /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf && \
     sed -i '/php_value session.save_path/d' /opt/rh/httpd24/root/etc/httpd/conf.d/php55-php.conf && \
     head -n151 /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf | tail -n1 | grep "AllowOverride All" || exit && \
-    mkdir /tmp/sessions && \
+    mkdir -p /tmp/sessions /opt/app-root/etc/conf.d && \
+    touch /opt/app-root/etc/conf.d/env.conf && \
+    ln -sf /opt/app-root/etc/conf.d/env.conf /opt/rh/httpd24/root/etc/httpd/conf.d/env.conf && \
+    chmod -R a+rwx /opt/app-root/etc/conf.d && \
     chmod -R a+rwx /opt/rh/php55/root/etc && \
     chmod -R a+rwx /opt/rh/httpd24/root/var/run/httpd && \
     chmod -R a+rwx /tmp/sessions && \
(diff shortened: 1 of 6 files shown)


remicollet:
Why all ENV would have to be available in Apache sub process ?

mfojtik:
@remicollet the original issue was about user trying to do:

```php
print "env is: ".$_ENV["USER"]."\n";
```

Which printed nothing. Then the user set a TESTVAR variable for container and that variable was empty as well. The mystery for me is how the other variables (like MYSQL one) get passed to PHP :-)

bparees:
Given we're already doing similar things elsewhere (https://github.com/openshift/sti-perl/blob/master/5.20/s2i/bin/run#L13-L15) i'm ok with this, but will wait for @remicollet to sign off or suggest an alternative as well.

mfojtik:
@bparees updated this PR to include 5.5 image + made it in sync with perl awk approach

bparees:
lgtm

mfojtik:
@remicollet you OK merging this?

hanzz:
I see the use-case here that sti-php user wants to be able to configure PHP app running in container using the ENV variables. I think that's quite easy way how to do that.

My concerns are that passing all env variables could mean passing also possible secret variables which could be later read by some 3rd person using some bug in PHP code.

I know there are [Secrets](https://docs.openshift.com/enterprise/3.0/dev_guide/secrets.html) and probably similar stuff in Openshift, so if you really believe all the people will use it with sti-php, I think it's safe to merge, otherwise I would think about different solutions to this problem.

remicollet:
Yes, not passing environment variables to apache modules and sub-process is a standard Apache feature (and reason for SetEnv directive)

So exposing "everything", without any filter, seems terrible strange.

I will definitively prefer "any" other solution.
At least, an option with the list of variable to expose this way (ex all variables set, and used by the "envsubst" command have no need to be exposed)

mfojtik:
@remicollet just brainstorming:

1) we can whitelist only env vars with HTTP_ prefix?
2) we can filter out all system env and pass only user set properties (that will be pretty hard to do)
3) we can have special env var where you list envs you want to pass (that seems terrible user experience)
4) we let you write your own env.conf (which I think is already possible to override with assemble script now).

hanzz:
I think allowing prefixed variables seems to be a good way. We can also show some kind of warning with a list of variables passed to PHP in the run script. Something like:

```
Following environment variables are passed to all PHP scripts executed by Apache httpd:
    -  ...
Make sure they are not containing sensitive data.
```

mfojtik:
@hanzz yes, I saw it. it is because awk does not filter these... I had for loop there but it was pretty ugly ;-)
Also we have already in sti-perl (assuming that image is broken now :)

mfojtik:
@hanzz @remicollet now what you suggest the prefix should be? `HTTP_` ? `PHP_` ?

bparees:
@mfojtik @hanzz @remicollet bump

remicollet:
I think adding new ENV for each need is going to be a nightmare...
I think PR #67 will have fix this issue and TZ one.

mfojtik:
@bparees I incline to close this PR till we figure out how we want to deliver envs to PHP. Also since there are no other bugs open for things like database envs I would prefer to wait till someone other than me complains :-)

PR #87, opened 2016-02-19
Set data.timezone in Dockerfile

tomkukral:
Assemble phase fails with composer applications due to unset _date.timezone_:

```
E0219 09:37:35.508703 1 util.go:91] Warning: date_default_timezone_get(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone.
```

This command returns 1 and thus assemble phase fails.

Can be reproduced with `s2i build https://github.com/tomaskadlec/d2s.git centos/php-56-centos7 obedar`

This PR adds sed command used for setting the timezone in Dockerfile. Tested only on `centos/php-56-centos7` was tested.

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -35,7 +35,6 @@ COPY ./contrib/ /opt/app-root
 # random UID.
 RUN sed -i -f /opt/app-root/etc/httpdconf.sed /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf && \
     sed -i '/php_value session.save_path/d' /opt/rh/httpd24/root/etc/httpd/conf.d/php55-php.conf && \
-    sed -i 's/^;date.timezone.*/date.timezone = GMT/g' /etc/opt/rh/rh-php56/php.ini && \
     head -n151 /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf | tail -n1 | grep "AllowOverride All" || exit && \
     mkdir /tmp/sessions && \
     chmod -R a+rwx /opt/rh/php55/root/etc && \
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -35,6 +35,7 @@ COPY ./contrib/ /opt/app-root
 # random UID.
 RUN sed -i -f /opt/app-root/etc/httpdconf.sed /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf && \
     sed -i '/php_value session.save_path/d' /opt/rh/httpd24/root/etc/httpd/conf.d/php55-php.conf && \
+    sed -i 's/^;date.timezone.*/date.timezone = GMT/g' /etc/opt/rh/rh-php56/php.ini && \
     head -n151 /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf | tail -n1 | grep "AllowOverride All" || exit && \
     mkdir /tmp/sessions && \
     chmod -R a+rwx /opt/rh/php55/root/etc && \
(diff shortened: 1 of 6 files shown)


mfojtik:
can you please squash?

LGTM

[test]

mfojtik:
also the test indicate that there might be problem with sed command.

mfojtik:
[test]

mfojtik:
@remicollet FYI

tomkukral:
I guest path `/etc/opt/rh/rh-php56/php.ini` is different in RHEL7

remicollet:
I think this "GMT" value is not relevant for most users.
Better to add something in the assemble  script.

mfojtik:
@tomkukral @remicollet something like `PHP_TZ` variable we then use to set the timezone? If the default (nothing is set) breaks users, then we should figure out some sane default. Perhaps default to what TZ is set in container?

tomkukral:
What about generating _php.ini_ file during assemble phase instead of the run phase?

mfojtik:
@tomkukral you can build the image in one TZ and run in different TZ :-)

tomkukral:
@mfojtik two different timezone vars ... `TZ` & `TZ_BUILD` seems like the only possible solution to me

tomkukral:
I'm suggesting different approach ... date.timezone is set only in case of composer is detected && timezone is not set. 
Can you please add RHEL suppoort and [test]? It should be sufficient to set **PHP_CONFIG** in _Dockerfile.rhel7_.

remicollet:
If you just need it for the composer run, just use: 

     php -d date.timezone=UTC   ./composer.phar install --no-interaction --no-ansi --optimize-autoloader

tomkukral:
Thanks for this idea @remicollet, it would be really great to solve it this way, but it doest not work. Composer runs another PHP processes and these totally ignore configuration provided by **-d**. 

It is probably caused the PHPExecuteFinder packed in composer and symfony. I'm not PHP guy, this idea comes from @tomaskadlec.

We ended with out custom build image and thus merging this PR is not so important for us, but I may help the others.

tomkukral:
Does this PR have a chance to get merged? Otherwise it would suggest to close it.

bparees:
@remicollet, @hhorak can you bring this to a conclusion?

hhorak:
[test]

hhorak:
@remicollet, what do you think, is this still worth fixing or is it too specific use case?

omron93:
@tomkukral In this PR some files are changed in 5.5 versions and some in 5.6. So changes are not complete and testing is failing... (`$PHP_CONFIG` is not set in 5.5 so `cp` fails)

remicollet:
Seems unneeded, see
https://github.com/sclorg/s2i-php-container/blob/master/5.6/contrib/etc/php.ini.template#L890

tomkukral:
@omron93 it is failing in RHEL too since I'm not able to test it. I'd be happy to finish this issue if anybody would be interested in this solution. We have switched to our custom images.

@remicollet seems, but it is needed. Referenced php.ini.templates is used during *run* phase (https://github.com/sclorg/s2i-php-container/blob/27c782cd7479cf5787ef97772875a9d6fec54998/5.6/s2i/bin/run#L24) and thus these configuration is not set in *assemble*.

I'd like to suggest to close this PR since nobody else need it.

omron93:
> it is failing in RHEL too since I'm not able to test it. I'd be happy to finish this issue if anybody would be interested in this solution. We have switched to our custom images.

@tomkukral You didn't changed `5.5/Dockerfile`. You should also change `5.5/Dockerfile.rhel7` and `5.6/Dockerfile.rhel7`.

hhorak:
Closing per request, let's fix it once someone else requires it as well.

PR #91, opened 2016-02-23
validate rpms actually get installed

bparees:


bparees:
@mfojtik eek. we weren't checking our php rpms and several weren't being installed.

@remicollet I had to remove a couple rpms that apparently don't exist (anymore?)

from centos php5.5: php55-php-pecl-imagick php55-php-pecl-xdebug
from centos php5.6: more-php56-php-pecl-imagick

also we're not installing imagick at all in the rhel php5.6 image.

Are there new package names for these?  I seem to recall something about imagemagick being part of another package in php5.6?

bparees:
the net here is:
centos php5.5 is missing imagick and xdebug
rhel php5.5 has both imagick and xdebug
centos php5.6 is missing imagick  (or that's part of another package now so we're fine)
rhel php5.6 is missing imagick (or that's part of another package now so we're fine)

bparees:
@hhorak @remicollet bump.

remicollet:
This have been already discussed.
The image only use package available in official RHSCL chanel.

So XDebug is only part of rh-php56 (and BTW is only for developers, and IMHO  should probably be removed for production usage).

Imagick is not part of RHSCL.

BTW, if you want a cleanup commit, please use 1 line per package, so future commit will be readable

mfojtik:
@remicollet this image can be used for development as well, so we might provide a way to enable "development" mode for PHP which in turn enable xdebug for users.

Are there any plans including ImageMagick in SCL for PHP? I do remember we had some use-cases for that and I think @tdawson was packaging that RPM for OpenShift v2.

remicollet:
> so we might provide a way to enable "development" mode for PHP which in turn enable xdebug for users.

Again, this was the goal of pr #67 

> Are there any plans including ImageMagick in SCL for PHP?

AFAIK: no

hhorak:
I don't see any open request for imagick, only this closed one: https://bugzilla.redhat.com/show_bug.cgi?id=1193530. We should track that if we should consider imagick for some further release.

bparees:
@remicollet 

> The image only use package available in official RHSCL chanel.
> Imagick is not part of RHSCL.

then what is "php55-php-pecl-imagick", which we are installing in the php55 RHEL image?

And why does there appear to be no centos version of it?

> So XDebug is only part of rh-php56 

so what is "php55-php-pecl-xdebug"?  And why is there no centos version of it?

> (and BTW is only for developers, and IMHO should probably be removed for production usage).

we don't have a way to distinguish production and developer usage in our images right now, so we need to include the superset of things required for either use case.

I'm sorry if these questions are tedious or redundant to previous discussions, but I don't live in the SCL world like you guys do, so it's harder for me to keep track of your packaging processes/rules/availability/etc, so I appreciate your patience.

remicollet:
@bparees I don't understand your question.

> then what is "php55-php-pecl-imagick", which we are installing in the php55 RHEL image?

No idea. IIRC It was originally pulled from "morephp55" repo,

> And why does there appear to be no centos version of it?

Don't understand what you mean.

php55 in RHSCL don't have imagick nor xdebug
php56 in RHSCL don't have imagick

This have already be discussed internally.
Most app have an optional dep. on imagick, because there is at least 3 extensiions for image: gd, imagick and gmagick. Only gd is provided in RHSCL.

bparees:
ok.  i'd love to know where we got the php55 imagick and xdebug packages from since apparently they've been packaged for rhel but not centos (perhaps @tdawson did that for us), but I guess going forward it doesn't matter that much.  We've got what we've got for php56 and i guess until/unless users express concerns over that, it should be sufficient.

mfojtik:
@bparees Troy was packaging them for us :-)

PR #92, opened 2016-03-17
Tune OPCACHE_MEMORY_CONSUMPTION based on available memory

mfojtik:


mfojtik:
@mnagy @bparees @remicollet PTAL

remicollet:
Sorry, but I don't think opcache have to be computed this way

It have to be tuned according to the served app. (number + size of files)
Having OOM opcache reset is terribly bad for perf (better to run without opcache)

Once again, I strongly believe a "generic" image just doesn't make sense.

mfojtik:
@remicollet we tuned it this way in v2. for generic image (docker run) there is no magic happening... this tuning is done only when you specify --memory or run on openshift with resource limits.

bparees:
> Having OOM opcache reset is terribly bad for perf (better to run without opcache)

sorry, what is an "OOM opcache reset" and what would cause it?  (i know OOM means out of memory).

remicollet:
@bparees sorry but I definitively don't understand the logic.

If you limit memory, you should first reduce process number, and probably memory_limit (used by each process), disable unneeded extensions.... before opcache size.

remicollet:
> what is an "OOM opcache reset" 

When the cache is full, opcache reset it and start with a new cache.

This is expected and not a issue if happen very few time (e.g. once per day) but can become a real issue if occurs too often (e.g. after only a few requests).

A cache MUST be tweak.

mfojtik:
@remicollet i'm not changing/tweaking the reset option

mfojtik:
@remicollet the goal here is to leverage the extra memory (iow. you run on openshift and give the php container 5GB of memory). In that case you expect the app will "use" that memory, so we want to tweak the cache size, etc. I'm open to suggestions about what other things we can "tweak" based on the amount of memory given to container.

5.5/s2i/bin/run
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Export the Cgroup limits
+export $(/usr/bin/cgroup-limits.py)
```
mnagy: bad, this will ignore errors and behave oddly if the script throws an exception.

Btw, should this run script use set -e?

5.6/s2i/bin/run
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Export the Cgroup limits
+export $(/usr/bin/cgroup-limits.py)
```
mnagy: same as above

5.5/s2i/bin/run
mfojtik: yes, it should be -e

mfojtik:
[test] !

bparees:
@mfojtik nothing will pass until we get a good AMI again :(  i'm working on it, but the AMI builds are super flaky right now.

5.5/s2i/bin/run
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# Export the Cgroup limits
+set -e
+export_vars=$(cgroup-limits) ; export $export_vars
+set +e
```
mnagy: You don't have to use set here just for one line. You could do this instead:
```
export_vars=$(cgroup-limits) || exit 1
export $export_vars
```
Alternatively, you could do:
```
export_vars=$(cgroup-limits) && export $export_vars || exit 1
```
which is shorter. I've tested it:
```
14:10 $ false && echo export || echo exit
exit
14:10 $ true && echo export || echo exit
export
```

5.6/s2i/bin/run
```diff
@@ -9,8 +14,17 @@ export TRACK_ERRORS=${TRACK_ERRORS:-OFF}
...
 export INCLUDE_PATH=${INCLUDE_PATH:-.:/opt/app-root/src:/opt/rh/rh-php56/root/usr/share/pear}
 export SESSION_PATH=${SESSION_PATH:-/tmp/sessions}
-# TODO should be dynamically calculated based on container memory limit/16
-export OPCACHE_MEMORY_CONSUMPTION=${OPCACHE_MEMORY_CONSUMPTION:-16M} 
+
+# In case the memory limit is set, use 10% of available memory for opcache memory
+# Users can always override the computed value using OPCACHE_MEMORY_CONSUMPTION
+# variable.
+if [[ "${MAX_MEMORY_LIMIT_IN_BYTES}" != "${MEMORY_LIMIT_IN_BYTES}" ]]; then
+  OPCACHE_MEMORY_COMPUTED="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.10))")M"
```
mnagy: You can get the same results with:
```
OPCACHE_MEMORY_COMPUTED="$((MEMORY_LIMIT_IN_BYTES/1024/1024/10))M"
```

5.5/s2i/bin/run
mfojtik: @mnagy i was following your example in cgroup-limits script ;-) but agree, I prefer || exit 1

mfojtik:
@mnagy comments addressed, thanks!

mfojtik:
[test]

mfojtik:
@bparees do we want this merged? I don't see anything else to tweak based on memory and we are tweaking the opcache in v2.

mfojtik:
[test]

bparees:
@mfojtik update the readme to describe this behavior?  And have you resolved the discussion with @remicollet ?

mfojtik:
closing in favor of https://github.com/openshift/s2i-php/pull/96

PR #94, opened 2016-04-21
send error log to stdout

bparees:
fixes https://github.com/openshift/s2i-php/issues/93

### DIFF
--- a/5.5/contrib/etc/httpdconf.sed
+++ b/5.5/contrib/etc/httpdconf.sed
@@ -4,6 +4,6 @@ s/^Group apache/Group root/
 s%^DocumentRoot "/opt/rh/httpd24/root/var/www/html"%DocumentRoot "/opt/app-root/src"%
 s%^<Directory "/opt/rh/httpd24/root/var/www/html"%<Directory "/opt/app-root/src"%
 s%^<Directory "/opt/rh/httpd24/root/var/html"%<Directory "/opt/app-root/src"%
-s%^ErrorLog "logs/error_log"%ErrorLog "/tmp/error_log"%
+s%^ErrorLog "logs/error_log"%ErrorLog "|/usr/bin/cat"%
 s%CustomLog "logs/access_log"%CustomLog "|/usr/bin/cat"%
 151s%AllowOverride None%AllowOverride All%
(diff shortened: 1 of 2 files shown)


bparees:
@ppisar @rhcarvalho ptal.

rhcarvalho:
Why not send to stderr? What was broken in Docker 1.8 (the comment in the issue)?

bparees:
@rhcarvalho see https://github.com/openshift/s2i-perl/issues/75#issuecomment-205829541

rhcarvalho:
@bparees ok got it. Isn't it a problem that the error and access logs get mixed together?

bparees:
it's not ideal, but it's the course of action we're taking elsewhere as well.  eg w/ the rest of the log subsystem, if you want your logs aggregated, they all have to go out to stdout for collection, so the assumption is you're going to find a way to demultiplex them if it matters to you.

rhcarvalho:
> the rest of the log subsystem

What does this mean? Everything 'oc logs', something else?

Given the limitations we have, changes LGTM.

bparees:
i meant the elastisearch/fluentd/kibana logging infrastructure.  but oc logs too since it doesn't give you a way to get the stderr logs.

i'm not even sure what docker does to let you see the stderr from a container?

ppisar:
On Thu, Apr 21, 2016 at 02:03:45PM -0700, Ben Parees wrote:
> i'm not even sure what docker does to let you see the stderr from
> a container?
> 
Provided a docker container is spawned by a docker daemon and docker client
can emmit container's stderr to client's stderr, the stderr stream must be
either transmitted from docker server to docker client via it's protocol or
the server duplicated the client's stderr descriptor to container process.

At the end, from point of view of a container process, it's just a file
descriptor inherited from a parent process. No magic.

-- Petr

bparees:
@ppisar right, i just meant whether the "docker logs" command gives you a way to differentiate between stderr and stdout from the container.  i'm not aware that it does.  it may well even merge them at the daemon side, so the client can't separate them (but again not sure).

rhcarvalho:
> i just meant whether the "docker logs" command gives you a way to differentiate between stderr and stdout from the container. i'm not aware that it does. it may well even merge them at the daemon side, so the client can't separate them (but again not sure).

@bparees Docker does keep them separate:

```console
$ docker run --entrypoint=/bin/bash centos -c 'echo 123; echo >&2 456' 2>/dev/null                                                                                 
123
$ docker run --entrypoint=/bin/bash centos -c 'echo 123; echo >&2 456' 1>/dev/null                                                                                 
456
$ docker ps -a -l
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS                      PORTS               NAMES
491909052a0c        28e524afdd05        "/bin/bash -c 'echo 1"   43 seconds ago      Exited (0) 42 seconds ago                       backstabbing_almeida
$ docker logs backstabbing_almeida
123
456
$ docker logs backstabbing_almeida 2>/dev/null
123
$ docker logs backstabbing_almeida 1>/dev/null                                                                                                                           
456
```

bparees:
@rhcarvalho ah, cool, thanks!

PR #95, opened 2016-04-28
Enable short_open_tags

spinolacastro:
@bparees Would be nice to have it enabled as 5.5 does. PTAL

### DIFF
--- a/5.5/s2i/bin/run
+++ b/5.5/s2i/bin/run
@@ -9,6 +9,7 @@ export TRACK_ERRORS=${TRACK_ERRORS:-OFF}
 export HTML_ERRORS=${HTML_ERRORS:-ON}
 export INCLUDE_PATH=${INCLUDE_PATH:-.:/opt/app-root/src:/opt/rh/php55/root/usr/share/pear}
 export SESSION_PATH=${SESSION_PATH:-/tmp/sessions}
+export SHORT_OPEN_TAG=${SHORT_OPEN_TAG:-ON}
 # TODO should be dynamically calculated based on container memory limit/16
 export OPCACHE_MEMORY_CONSUMPTION=${OPCACHE_MEMORY_CONSUMPTION:-16M} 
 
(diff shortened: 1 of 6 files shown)


bparees:
would it make sense for this to be configurable?

bparees:
@remicollet any idea why we changed the default on this in 5.6?

remicollet:
No idea if (and why) this have changed from previous image, but [short tags](http://php.net/manual/en/ini.core.php#ini.short-open-tag) are deprecated and discouraged for year, especially because of conflict with <?xml tag.

I really think we should keep them disabled by default.

remicollet:
Notice: short tags are also disabled by default in RHEL since PHP 5.3 (php53 in RHEL-5 or php in RHEL-6)

bparees:
thanks @remicollet 
@spinolacastro with that in mind, i think we'd only want to accept this if you instead make it a configurable option (via env var like the other options) that defaults off.

spinolacastro:
Thanks, i'll make the changes.

2016-04-29 10:25 GMT-03:00 Ben Parees <notifications@github.com>:

spinolacastro:
@bparees PTAL

5.5/s2i/bin/run
```diff
@@ -9,6 +9,7 @@ export TRACK_ERRORS=${TRACK_ERRORS:-OFF}
 export HTML_ERRORS=${HTML_ERRORS:-ON}
 export INCLUDE_PATH=${INCLUDE_PATH:-.:/opt/app-root/src:/opt/rh/php55/root/usr/share/pear}
 export SESSION_PATH=${SESSION_PATH:-/tmp/sessions}
+export SHORT_OPEN_TAG=${SHORT_OPEN_TAG:-OFF}
```
bparees: we should keep the 5.5 behavior as it was, lest we break someone's app, so default this to On.

bparees:
@spinolacastro please add this to the readmes for 5.5 and 5.6:
https://github.com/openshift/s2i-php/blob/master/5.6/README.md#environment-variables
https://github.com/openshift/s2i-php/blob/master/5.5/README.md#environment-variables

spinolacastro:
@bparees done

bparees:
lgtm [test]

PR #99, opened 2016-05-16
[WIP] Synchronize MPM tuning conf file with Perl's

mnagy:
MaxRequestsPerChild has been renamed to MaxConnectionsPerChild in
Apache 2.4.

I was also thinking about limiting the `HTTPD_MAX_REQUEST_WORKERS` as done in the Perl PR. I'm a bit afraid of setting it, by accident, to a much higher value than would be appropriate. We also do not allow it to be set explicitly (if you set a limit, it will be overwritten).

@mfojtik @bparees, thoughts? I'd like to add those changes as proposed into this PR, but wanted to consult before doing the work. See here for the code I'd like to use: https://github.com/mnagy/sti-perl/blob/7f4ffecbf0efda7d4c13041581bf55e8bd2786d4/5.16/s2i/bin/run#L20-L36

### DIFF
--- a/5.5/contrib/etc/conf.d/50-mpm-tuning.conf.template
+++ b/5.5/contrib/etc/conf.d/50-mpm-tuning.conf.template
@@ -1,12 +1,13 @@
 <IfModule mpm_prefork_module>
     # This value should mirror what is set in MinSpareServers.
-    StartServers          ${HTTPD_START_SERVERS}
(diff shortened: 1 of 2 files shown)


bparees:
sounds reasonable to me.

bparees:
@remicollet ptal

bparees:
nm, looks like we must have already picked up this change elsewhere.

PR #109, opened 2016-07-01
Use OSE-3.2 repo, change sti base image

pkubatrh:


5.5/Dockerfile.rhel7
```diff
@@ -1,4 +1,4 @@
-FROM openshift/base-rhel7
+FROM openshift3/sti-base:1.0
```
bparees: this isn't a valid change.  the 5.5 image is built via a different process than the 5.6 image because it's owned by the openshift team, not the SCL team, and we have our own base-rhel7 image that we use.

(and openshift3/sti-base isn't an image)

hhorak:
[test]

Issue #107, opened 2016-06-29
Configurable documentroot?


andrewklau:
It would be desirable if we could have the DocumentRoot configurable as an environment variable, ie. our repository is configured to have composer.json in the root directory, but a different directory stores the public content (ie. /web or /public)

andrewklau:
Perhaps we could try define something like this?

```
<IfDefine pub>
  Define documentroot /pub
</IfDefine>
<IfDefine web>
  Define documentroot /web
</IfDefine>
DocumentRoot "/opt/app-root/src${documentroot}"
```

changing the run to:
`httpd -D frontend -D ${DOCUMENTROOT}`

hhorak:
Fixed by #111.

PR #111, opened 2016-07-02
Adding DOCUMENTROOT environment variable

andrewklau:
Much simpler then my original concept #107

### DIFF
--- a/5.5/contrib/etc/httpdconf.sed
+++ b/5.5/contrib/etc/httpdconf.sed
@@ -1,7 +1,7 @@
 s/^Listen 80/Listen 0.0.0.0:8080/
 s/^User apache/User default/
 s/^Group apache/Group root/
-s%^DocumentRoot "/opt/rh/httpd24/root/var/www/html"%DocumentRoot "/opt/app-root/src"%
+s%^DocumentRoot "/opt/rh/httpd24/root/var/www/html"%#DocumentRoot "/opt/app-root/src"%
 s%^<Directory "/opt/rh/httpd24/root/var/www/html"%<Directory "/opt/app-root/src"%
 s%^<Directory "/opt/rh/httpd24/root/var/html"%<Directory "/opt/app-root/src"%
 s%^ErrorLog "logs/error_log"%ErrorLog "|/usr/bin/cat"%
--- a/5.5/README.md
+++ b/5.5/README.md
@@ -109,6 +109,9 @@ The following environment variables set their equivalent property value in the p
 * **SHORT_OPEN_TAG**
   * Determines whether or not PHP will recognize code between <? and ?> tags
   * Default: ON
+* **DOCUMENTROOT**
+  * Path that defines the DocumentRoot for your application (ie. /public)
+  * Default: /
 
 The following environment variables set their equivalent property value in the opcache.ini file:
 * **OPCACHE_MEMORY_CONSUMPTION**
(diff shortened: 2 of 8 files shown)


bparees:
[test]

bparees:
@remicollet ptal

remicollet:
We have in httpd.conf
`echo "IncludeOptional /opt/app-root/etc/conf.d/*.conf"`

Isn't this enough to override this configuration?
(else this recently added include doesn't make sense)

andrewklau:
@remicollet Aren't I already doing that with this `envsubst < /opt/app-root/etc/conf.d/00-documentroot.conf.template > /opt/app-root/etc/conf.d/00-documentroot.conf` similar to the 50-mpm-tuning.conf ?

remicollet:
@andrewklau yes, we can continue to add environment variable for each httpd/php directive.

And the whole result will become a mess.
The reason why I really think we need a global solution.

andrewklau:
@remicollet anyway I could help to look into a global solution? Not having the ability to set the documentroot blocks us from being able to make this transition

andrewklau:
@bparees @remicollet bump

5.5/contrib/etc/httpdconf.sed
```diff
@@ -1,7 +1,7 @@
 s/^Listen 80/Listen 0.0.0.0:8080/
 s/^User apache/User default/
 s/^Group apache/Group root/
-s%^DocumentRoot "/opt/rh/httpd24/root/var/www/html"%DocumentRoot "/opt/app-root/src"%
+s%^DocumentRoot "/opt/rh/httpd24/root/var/www/html"%#DocumentRoot "/opt/app-root/src"%
```
bparees: is the added # intentional?

andrewklau: @bparees yep, because I add the DocumentRoot by the template 00-documentroot.conf.template

`DocumentRoot "/opt/app-root/src${DOCUMENTROOT}"`

bparees:
>  We have in httpd.conf
> echo "IncludeOptional /opt/app-root/etc/conf.d/*.conf"
> Isn't this enough to override this configuration?
> (else this recently added include doesn't make sense)

How will the user inject a file into that directory?  the assemble script is going to move all the provided files into /opt/app-root/src.

the new includeoptional was done to pick up the new tuning file which is created at runtime, so @andrewklau is following that pattern.

In any case, regardless of future plans to handle config differently (at which point this logic can be removed), I don't see a good reason to hold up this PR in the short term.

@hhorak ?

5.5/contrib/etc/httpdconf.sed
bparees: sorry, it's too late here for me to be parsing regex with non-standard delimiters, i see what you're doing now. (commenting the line out).

hhorak:
> How will the user inject a file into that directory? the assemble script is going to move all the provided files into /opt/app-root/src.

I think assemble should use similar way as httpd container does it (https://github.com/sclorg/httpd-container/blob/master/2.4/s2i/bin/assemble#L20 or build php container on top of httpd one finally).

> In any case, regardless of future plans to handle config differently (at which point this logic can be removed), I don't see a good reason to hold up this PR in the short term.

I agree, unless there is something ready to be used, we simply need to go with imperfectness.

LGTM.

bparees:
> I think assemble should use similar way as httpd container 

yeah it probably should, i was just making the point that today it doesn't, so a user can't do that w/o changing the assemble script.

Issue #106, opened 2016-06-15
Assemble script cleanup source


spinolacastro:
Hello, i realized that images builded by sti has a considerable size, looking for a answer i've found the source pulled from a git repository is located twice on the output image, once at `/opt/app-root/src` and  `/tmp/src`.

Am i missing something or we can add `rm -rf /tmp/src` at the bottom of assemble script?

bparees:
you can, or you can change the assemble script to do mv instead of cp, this has been discussed here as well:
https://github.com/sclorg/s2i-python-container/issues/84

as you can see there, there are some subtleties to it, as discussed here:
https://github.com/sclorg/s2i-python-container/issues/84#issuecomment-224471693

i'd be open to a PR to the assemble script though, that used a mv instead of cp.

bparees:
assigning myself for now to keep this on my radar, but if you're interested in doing the PR let me know.

PR #114, opened 2016-07-12
Use mv instead cp to save space

spinolacastro:
@bparees this have been used on 5.5 and 5.6 tags without issues, PTAL.

fixes https://github.com/sclorg/s2i-php-container/issues/106

### DIFF
--- a/5.5/s2i/bin/assemble
+++ b/5.5/s2i/bin/assemble
@@ -2,8 +2,9 @@
 
 set -e
 
+shopt -s dotglob
 echo "---> Installing application source..."
-cp -Rf /tmp/src/. ./
+mv /tmp/src/* ./
 
 if [ -f composer.json ]; then
   echo "Found 'composer.json', installing dependencies using composer.phar... "
--- a/5.6/s2i/bin/assemble
+++ b/5.6/s2i/bin/assemble
@@ -2,8 +2,9 @@
 
 set -e
 
+shopt -s dotglob
 echo "---> Installing application source..."
-cp -Rf /tmp/src/. ./
+mv /tmp/src/* ./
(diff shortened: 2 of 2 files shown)


bparees:
[test]

5.5/s2i/bin/assemble
```diff
@@ -2,8 +2,9 @@
 
 set -e
 
+shopt -s globstar
 echo "---> Installing application source..."
-cp -Rf /tmp/src/. ./
+mv /tmp/src/* ./
```
bparees: i think you need dotglob or you're going to miss dot files.

spinolacastro: Isn't  `shopt -s globstar` supposed to do that?

bparees: not based on my reading and not based on my testing:
```
ls -a
.  ..  .bar  .foo  foo2  .s2i
```

```
$ shopt -s globstar
$ ls *
foo2
```

```
$ shopt -s dotglob
$ ls *
.foo  foo2

.bar:
foo2

.s2i:
bin
```

spinolacastro: Sorry, my bad.

5.5/s2i/bin/assemble
```diff
@@ -2,8 +2,9 @@
 
 set -e
 
+shopt -s globstar dotglob
```
bparees: globstar isn't needed/doing anything
http://www.linuxjournal.com/content/globstar-new-bash-globbing-option

bparees:
[test]

bparees:
lgtm

PR #117, opened 2016-08-05
set PHP date.timezone at assebly if not set

arska:
Similar to #87 I had the problem that composer needs the timezone to be set at assemble-time while we set it at run-time.

This patch sets the timezone if composer needs to run and the user has not set it separately

### DIFF
--- a/5.5/s2i/bin/assemble
+++ b/5.5/s2i/bin/assemble
@@ -9,6 +9,12 @@ mv /tmp/src/* ./
 if [ -f composer.json ]; then
   echo "Found 'composer.json', installing dependencies using composer.phar... "
 
+  PHP_CONFIG=/opt/rh/php55/root/etc/php.ini
+  if [ "$(grep '^data\.timezone' $PHP_CONFIG ; echo $?)" -eq "1" ] ; then
+    # composer et al need timezone defined, php.ini will be generated at runtime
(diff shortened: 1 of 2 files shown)


bparees:
[test]

arska:
FYI this and #111 was needed to get symfony/symfony-demo to run on OpenShift Enterprise 3.2: https://github.com/arska/symfony-demo/commit/a12ab80328fd62d38baa161909c8826af47dd846

http://symfony-demo-symfony-demo.app.appuio-beta.ch/

hhorak:
This makes sense to me. @remicollet, are you fine with this?

remicollet:
@arska please explain how it breaks.

AFAIK, composer already set it to UTC (when needed)

See https://github.com/composer/composer/blob/master/src/Composer/Console/Application.php#L72

arska:
Hello @remicollet 

The problem in my case isn't composer itself but post-install-cmd scripts that are started by composer - in this case the Symfony ones at https://github.com/symfony/symfony-demo/blob/master/composer.json#L38 that fail without a timezone

Cheers,
Aarno

bparees:
@remicollet bump

remicollet:
Unneeded, see https://github.com/sclorg/s2i-php-container/blob/master/5.6/contrib/etc/php.ini.template#L890

bparees:
thanks @remicollet

PR #119, opened 2016-08-10
add jenkins badge

PI-Victor:
@hhorak cc


bparees:
Also you should check your workflow because these PRs all have incorrect commit descriptions.

PR #124, opened 2016-10-06
Added cccp.yml files for centos container pipeline.

mohammedzee1000:
This is for the purpose of building these container
on the CentOS Container pipeline.


hhorak:
It does not influence image functionality, so no need to test. LGTM.

PR #104, opened 2016-05-31
Curl instead of php readfile

Rydekull:
Use curl instead of php readfile to honor proxy settings.

PR #125, opened 2016-10-19
Curl instead of php readfile

bparees:


bparees:
@hhorak @remicollet ptal.

remicollet:
I don't understand "why", but I don't mind if you want it.

remicollet:
BTW, upstream way : 
```
php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
php -r "if (hash_file('SHA384', 'composer-setup.php') === 'e115a8dc7871f15d853148a7fbac7da27d6c0030b848d9b3dc09e2a0388afed865e6a3d6b3c0fad45c48e2b5fc1196ae') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
php composer-setup.php
php -r "unlink('composer-setup.php');"
```

bparees:
@remicollet sorry i left out the history from the PR this was copied from.  basically using curl means we can respect http proxy settings when pulling the file (https://github.com/sclorg/s2i-php-container/pull/104#issue-157635995)

as for what upstream is doing, that seems incredibly fragile...  if that hash ever changed, our image would be broken until we updated it.

PR #126, opened 2016-10-22
Updated repository name to sclorg/s2i-php-container in readme files

jkramarz:


bparees:
[test]

jkramarz:
Can't comment on breaking RHEL7 build because log is not public ;-)

PR #128, opened 2016-10-25
Add php-gmp module

hug-abacus:
Please add the php gmp module to the s2i image.


hhorak:
Sounds fine to you, @remicollet ?

remicollet:
Yes, lgtm (+ same in php 7.0)

remicollet:
BTW, perhaps we should not accept any change for PHP 5.5 which is EOL
Only 5.6 and 7.0 are now maintained

hug-abacus:
@remicollet is php7.0 in a different repo?

remicollet:
No, see PR #134

PR #133, opened 2016-11-12
modify variables_order to populate $_ENV

### LINKED ISSUE #83: The environment variables are not available for PHP
I think this problem existed in v2 as well. The environment variables has to be listed in the appropriate config in Apache, otherwise PHP won't be able to see them.

sjenning:
Fixes #83


bparees:
[test]

bparees:
lgtm, @remicollet ?

bparees:
(I should note there are security reasons why this might not be a good default, but seemingly we were ok with it in 5.5)

remicollet:
lgtm

PR #135, opened 2016-11-23
Fix path in README.md

hhorak:


hhorak:
Easy change in README, I don't think we need to wait for test results or review.

omron93:
To note: CentOS CI failed due to https://github.com/sclorg/rhscl-container-ci/issues/10

Issue #139, opened 2016-12-19
php 5.6 performance issues


andrewklau:
We have been receiving increased reports for performance issues with the 5.6 image.

5.5 and 7.0 work without any issues

bparees:
can you elaborate?  what sort of performance issue?  memory usage? cpu?  during build? application runtime?  has it gone up from previous versions of the 5.6 image?  how does it compare with the same metrics for the 5.5 image?

andrewklau:
Sorry edited my initial message just as you replied.

Applications running with 5.6 are significantly slower to respond which often leads to 504 responses
Switching to use 5.5 or 7.0 do not have the problem and they perform much faster

bparees:
Can you define significantly?  does this apply to trivial(hello world) apps?  Have you done any profiling to see where the time is being spent?

andrewklau:
The current issue is reported based on prod running apps like WordPress and
Laravel, returning the 504 timeout.

Switching to 5.5 and 7.0 resulted in massive improvements. Eg. One WP site
used to take 10+ seconds to render after receiving 2 simultaneous users.
Now it's almost under a second.

Im investigating further now. Will report back with other findings.

andrewklau:
Here is just some basic tests a customer ran, this seems to run slower across all the PHP images (in comparison to local PHP and an OpenShift 2 cluster):

time php test.php

```
<?php

function tryAlloc($megabyte){ 
echo "try allocating {$megabyte} megabyte..."; 
$mark = microtime(true); 
$dummy = str_repeat("-",1048576*$megabyte); 
echo "allocation time : " . sprintf("%f", microtime(true)-$mark) . "\n"; 
echo "Usage: " . memory_get_usage(true)/1048576; 
echo " Peak: " . memory_get_peak_usage(true)/1048576; 
echo "\n"; 
}

tryAlloc(1); 
tryAlloc(10); 
tryAlloc(20); 
tryAlloc(30); 
tryAlloc(40); 
tryAlloc(50); 
tryAlloc(100);
```

php-56-centos7: 

```
real	0m0.659s
user	0m0.031s
sys	0m0.324s
```

php-70-centos7:

```
real	0m0.136s
user	0m0.039s
sys	0m0.082s
```

php 5.4 (openshift 2):

```
real	0m0.244s 
user	0m0.056s 
sys	0m0.179s
```

PHP 7.0 (centos 7 own docker image):

```
real	0m0.057s
user	0m0.029s
[...]

PR #144, opened 2016-12-22
Remove xdebug from PHP 5.6

### LINKED ISSUE #122: disable xdebug option
It seems strange to be including xdebug for prod containers, it'd be ideal if we could add an option to disable this by ENV variable (or not include it at all?)

andrewklau:
Fixes #122 and #139 

@joellarson is right about memcache however this might potentially effect some users(?)


andrewklau:
@bparees ping

JoelLarson:
It would be totally fine to have both memcache and memcached installed, though the memcached extension does everything the memcache extension does.

bparees:
this is up to @hhorak and team at this point, but i'm definitely wary of the possibility that we break users who's apps expect this module is installed.  It might be better to just tell people to turn it off if they care.

sebas5384:
XDebug could be enabled optionally, not by default. Installed by default isn't the problem in my opinion.

PR #147, opened 2017-01-16
Log the correct remote Host IP

mhutter:
* Log `%{X-Forwarded-For}i` instead of `%h`


hhorak:
@remicollet Please, take a look.

hhorak:
[test]

remicollet:
The syntax seems correct, but I don't see why this is needed, I'm not aware of docker setting the X-Forwarded-For header.

notroj:
This should be configurable rather than done by default, it's specific to the environment in which the container is expected to run.

mhutter:
Hmm true. However I think it's possible to configure apache to log `X-Forwarded-For` if present, and `%h` otherwise. This way both setups would work out of the box.

I'll check

notroj:
Possibly the ideal way would be to log %a by default and then the s2i run script can set RemoteIP* based off some env var like HTTPD_TRUSTED_FORWARDERS being set.

PR #153, opened 2017-03-16
Set default opcache.memory_consumption to 128MB

### LINKED ISSUE #152: Opache memory consumption default too low for PHP7
The Zend Opcache memory consumption default of 16MB is too low in a PHP7 environment with interned string buffer set to 8MB (currently non-configurable in 10-opcache.ini).

This issue is described in https://bugs.php.net/bug.php?id=72720 

Why was 16MB chosen as the default for this container? The default set in PHP documentation for Opcache is 64MB and the scl-php70 package sets this at 128MB. Should the s2i-php container use this same default or another value? Small applications will happily run in 32MB allocated whereas Wordpress and Drupal will prefer 64 or 128.

Prior to #151, OpCache was allocated 128MB due to the run script not setting the container template correctly. Perhaps this value should be used as the default for the container to ensure BC.

lucasnetau:
Resolve #152 by setting default for Opcache Memory Consumption to 128MB on PHP5.6 and PHP7.0

### DIFF
--- a/5.6/README.md
+++ b/5.6/README.md
@@ -116,7 +116,7 @@ The following environment variables set their equivalent property value in the p
 The following environment variables set their equivalent property value in the opcache.ini file:
 * **OPCACHE_MEMORY_CONSUMPTION**
   * The OPcache shared memory storage size
-  * Default: 16M
+  * Default: 128M
 * **OPCACHE_REVALIDATE_FREQ**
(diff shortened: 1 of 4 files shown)


omron93:
[test]

omron93:
@lucasnetau Thanks.

Changing opcache default size was aggreed in #152 so merging this PR.

PR #164, opened 2017-08-23
Add incremental builds

### LINKED ISSUE #145: Implement save-artifacts to speed up composer
Are there plans to implement the `save-artifacts` into the other containers so we can speed up build times?

blesenechal:
Fixes #145


5.6/s2i/bin/save-artifacts
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+pushd ${HOME} >/dev/null
+tar cf - vendor
+popd >/dev/null
```
omron93: Git is complaining there about "No newline at end of fine". Please add it to `save-artifacts` files for all versions. Thanks.

omron93:
[test]

blesenechal:
@omron93 : Fixed

omron93:
Thanks.
It looks good to me, but I'm not php expert and I'm not really sure how composer works (what is `vendor` directory for).

@remicollet Please take a look.

remicollet:
@omron93 I have no idea what is the goal of this ;)  (what is save_artefacts goal)

BTW "vendor" is the directory where all dependencies of the projects are download during "composer install")

remicollet:
Notice: no value fix the 5.5 tree, the SCL is not EOL / unmaintained, should even probably be dropped.

omron93:
> @omron93 I have no idea what is the goal of this ;) (what is save_artefacts goal)

More info there - https://github.com/openshift/source-to-image#incremental-builds . `safe-artifacts` is used to speed up builds by using some "data" from previous build.

> BTW "vendor" is the directory where all dependencies of the projects are download during "composer install")

Thanks. So if it makes sense to me to copy them then this PR looks good to me.

> Notice: no value to fix the 5.5 tree, the SCL is EOL / unmaintained, should even probably be dropped.

@blesenechal That is right. Please drop changes from 5.5 version.

andrewklau:
vendor will have the composer dependencies. There are some caveats like sometimes it's actually slower (depending on your registry speed or previous image size).

If a dependency is removed from `composer.json` it will persist in the image forever. Perhaps need a clean artifacts variable or something.

blesenechal:
@omron93 : I've reverted the changes to the 5.5 branch.

> If a dependency is removed from composer.json it will persist in the image forever. Perhaps need a clean artifacts variable or something.

In fact, on the next `composer install`, the unused dependancies will be removed.

omron93:
@blesenechal Thanks for the fast fix. Last think - we are using shared scripts for building (git submodule). There was an update recently so CI was update according to it so that is why it failed.
Please run `git submodule update --remote` inside your `s2i-php-contianer` copy -> `common/` directory should be updated -> commit this change and push it into this PR. Thank you and sorry for troubles.

> There are some caveats like sometimes it's actually slower (depending on your registry speed or previous image size).

I hope no one is forced to use incremental builds. So I think this shouldn't cause any troubles.

blesenechal:
@omron93 Common submudule updated

omron93:
[test]

omron93:
Tests are passing, so I don't see any reason why not to merge. So merging.

@blesenechal Thank you for your contribution.

PR #166, opened 2017-08-23
Add composer args to the builder

### LINKED ISSUE #165: Allow installing without dev packages
The `assemble` script executes the following Composer command :
> ./composer.phar install --no-interaction --no-ansi --optimize-autoloader

By default, Composer installs the dev packages (defined in `require-dev` in `composer.json`) which is not always needed for production deployments (it increases the build time, generates bigger images, generates a bigger classes autoloaded which decreases performance, install extra packages with potential security issues, ...)

It would be nice to have a build option to execute Composer with the `--no-dev` option.

I'll try to create a pull request soon.

blesenechal:
Fixes #165, adding the `COMPOSER_ARGS = --no-dev` environment variable to the build with prevent composer from installing dev dependancies.

This pull request is a bit more generic as it allows to define any composer parameter through the `COMPOSER_ARGS` environment variable.

If `COMPOSER_ARGS` is not set or empty, the behaviour is the same as the previous behaviour, so no backward compatibily break.


remicollet:
Make sense.

Need to be rebased + 7.1 is missing

blesenechal:
@remicollet I've rebased on master and added PHP 7.1

5.6/README.md
```diff
@@ -130,6 +130,8 @@ yourself:
       * Adds a custom composer repository mirror URL to composer configuration. Note: This only affects packages listed in composer.json.
     * **COMPOSER_INSTALLER**
       * Overrides the default URL for downloading Composer of https://getcomposer.org/installer. Useful in disconnected environments.
+    * **COMPOSER_ARGS**
+      * Adds extra arguments to the `composer install` command line (for example (`--no-dev`).
```
omron93: Second `(` should be removed?
(in all versions)

blesenechal:
@omron93 Fixed

PR #169, opened 2017-09-15
Add mongodb driver

wgordon17:
Signed-off-by: Will Gordon <wgordon@redhat.com>

### DIFF
--- a/7.0/Dockerfile
+++ b/7.0/Dockerfile
@@ -35,7 +35,7 @@ RUN yum install -y centos-release-scl && \
     INSTALL_PKGS="rh-php70 rh-php70-php rh-php70-php-mysqlnd rh-php70-php-pgsql rh-php70-php-bcmath \
                   rh-php70-php-gd rh-php70-php-intl rh-php70-php-ldap rh-php70-php-mbstring rh-php70-php-pdo \
                   rh-php70-php-process rh-php70-php-soap rh-php70-php-opcache rh-php70-php-xml \
-                  rh-php70-php-gmp" && \
+                  rh-php70-php-gmp sclo-php70-php-pecl-mongodb" && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS --nogpgcheck && \
     rpm -V $INSTALL_PKGS && \
     yum clean all -y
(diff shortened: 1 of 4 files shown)


7.0/Dockerfile
```diff
@@ -55,6 +55,8 @@ RUN sed -i -f /opt/app-root/etc/httpdconf.sed /opt/rh/httpd24/root/etc/httpd/con
     echo "IncludeOptional /opt/app-root/etc/conf.d/*.conf" >> /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf && \
     mkdir /tmp/sessions && \
     chown -R 1001:0 /opt/app-root /tmp/sessions && \
+    #chown -R 1001:0 `pecl config-get php_dir`/.. && \
+    #chmod -R ug+rwx `pecl config-get php_dir`/.. && \
```
omron93: Why is this needed?

omron93:
Thank you for this PR.

@hhorak What about installing community SCL packages? What about a support for this?

@notroj @remicollet What do you think?

[test]

remicollet:
> What about installing community SCL packages? What about a support for this?

This will make image diverge from RH official image.

7.0/Dockerfile
wgordon17: Apologies, it's not needed at all

wgordon17:
@remicollet Does updating the Dockerfile.rhel7 files solve your reservation?

remicollet:
@wgordon17 no, as the mongodb extension is not available in RHSCL.

hhorak:
I don't think we should diverge between centos and rhel images, moving from community images (centos) to supported images (rhel) should not require any changes. I'd rather think about whether there is another way how to install the the extension. In case of ruby for example, all such missing connectors are installed as any other missing dependency, from upstream repository. Is something like that possible for php?

remicollet:
For pur-PHP library, can be installed using "composer" (already managed by the image)
For C extension, can be install using "pecl", btw, this requires a C compiler, and devel packages

     pecl install mongodb-1.3.0

hhorak:
@wgordon17, would installing by composer be acceptable solution for you?

remicollet:
@hhorak mongodb is a C extension (so s/composer/pecl/)

wgordon17:
@remicollet @hhorak `pecl install mongodb` was my first thought, and in trying to figure out permissions, I just assumed a `yum install` would've been easier. But I can certainly revisit the permissions needed for `pecl install` to work.

omron93:
@wgordon17 Any progress in this?

It seems that  @maicolcantagallo is facing same problem in https://github.com/sclorg/s2i-php-container/issues/174

wgordon17:
I have not had any progress yet. Apologies for the delay.


From: centos-ci <notifications@github.com> <notifications@github.com>
Reply: sclorg/s2i-php-container
<reply@reply.github.com>
<reply@reply.github.com>
Date: November 28, 2017 at 10:59:18
To: sclorg/s2i-php-container <s2i-php-container@noreply.github.com>
<s2i-php-container@noreply.github.com>
CC: Will Gordon <wgordon@redhat.com> <wgordon@redhat.com>, Mention
<mention@noreply.github.com> <mention@noreply.github.com>
Subject:  Re: [sclorg/s2i-php-container] Add mongodb driver (#169)

Can one of the admins verify this patch?

phracek:
This ticket can be closed. MongoDB is not supported at all. Feel free to reopen it.

PR #170, opened 2017-09-20
Enable rh-nodejs6, so users can use npm for installing JavaScript stuff

hhorak:
Similarly to https://github.com/sclorg/s2i-python-container/pull/209


remicollet:
I don't understand how this is expected to work...
Enabling the collection without installing it ?

pkubatrh:
>I don't understand how this is expected to work...
Enabling the collection without installing it ?

It is already installed via:
https://github.com/sclorg/s2i-base-container/blob/master/base/Dockerfile#L42

PR #175, opened 2017-10-12
Keep version and relese synced with downstream

pkubatrh:


omron93:
Thanks. LGTM, so merging.

PR #178, opened 2017-11-12
Add execution permission to save-artifacts

itscaro:
Currently the save-artifacts cannot be run by the shell and it provokes:

```
/usr/bin/container-entrypoint: line 2: /usr/libexec/s2i/save-artifacts: Permission denied
/usr/bin/container-entrypoint: line 2: exec: /usr/libexec/s2i/save-artifacts: cannot execute: Permission denied
```


omron93:
[test]

omron93:
This makes sense to me. @itscaro Thanks for the fix.

Tests are passing, so merging.

PR #179, opened 2017-11-20
tests: replace removed s2i flag

pkubatrh:
Similarily to: https://github.com/sclorg/s2i-python-container/pull/229


omron93:
Thanks for the fix.

Tests are passing. Merging.

PR #182, opened 2017-12-04
retry https://getcomposer.org/installer download if it fails

jim-minter:
We're getting periodic test flakes on Origin due to failure to download https://getcomposer.org/installer, it would be useful to see if a retry helps
https://github.com/openshift/origin/issues/16772
@bparees

### DIFF
--- a/5.5/s2i/bin/assemble
+++ b/5.5/s2i/bin/assemble
@@ -10,7 +10,18 @@ if [ -f composer.json ]; then
   echo "Found 'composer.json', installing dependencies using composer.phar... "
 
   # Install Composer
-  curl https://getcomposer.org/installer | php
+  TEMPFILE=$(mktemp)
+  RETRIES=6
+  for ((i=0; i<$RETRIES; i++)); do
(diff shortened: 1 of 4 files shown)


bparees:
[test]

omron93:
[test]

LGTM, thanks for the fix.

@bparees We was forced to do a change in CentOS CI - to be able to trigger it you have to be **public** member of sclorg/ github org.

omron93:
@jim-minter Why not to use `curl --retry 5 --retry-delay 10`?

jim-minter:
@omron93 according to the curl docs, `--retry` retries on a "transient error" ("a timeout, an FTP 4xx response code or an HTTP 5xx response code").  It's not clear to me that the error we're seeing fits into any of those categories.

omron93:
> according to the curl docs, --retry retries on a "transient error" ("a timeout, an FTP 4xx response code or an HTTP 5xx response code"). It's not clear to me that the error we're seeing fits into any of those categories.

Thanks for explanation.

omron93:
@jim-minter Thanks for this fix.

PR #183, opened 2017-12-04
Enable SSL support and refactor image sources

hhorak:
This PR is a bit bigger than I originally hoped. There are basically these issues addressed here:

1) The most significant change is adding SSL feature -- by installing the module.

2) However, when running SSL server, user should be able to provide own SSL certificates, which is technically extending the image, so the general extendibility was also added as part of this PR, similar to what was already done in case of httpd container (https://github.com/sclorg/httpd-container/).

3) When working with configuration files, it turned to be pain to do it when those paths are hard-coded in scripts, while it also prevents it from sharing the same code across versions and platforms. Thus, I decided to also replace the hard-coded paths by variables defined in the Dockerfile. It allows to share the scripts better, makes it less error-prone for copy-paste errors and also allows to share scripts with Fedora based image.

4) As a proof for the last point, Fedora image is attached as part of this PR as well.


hhorak:
This addresses https://bugzilla.redhat.com/show_bug.cgi?id=1478556 btw.

hhorak:
@remicollet WDYT?

5.6/root/usr/share/container-scripts/php/common.sh
```diff
@@ -0,0 +1,140 @@
...
+  fi
+}
+
+# Copy SSL files provided in application source
+# This function was taken from httpd container
+process_ssl_certs() {
+  local dir=${1:-.}
+  if [ -d ${dir}/httpd-ssl/private ] && [ -d ${dir}/httpd-ssl/certs ]; then
+    echo "---> Looking for SSL certs for httpd..."
+    cp -r ${dir}/httpd-ssl ${HTTPD_APP_ROOT}
```
omron93: `HTTPD_APP_ROOT` is missing. (cause of CI failure)

hhorak: Thx, it's was a copy/paste issue, fixing it just now.

remicollet:
LGTM (quick look), as far as synced with httpd container change which have been merged.

hhorak:
[test-openshift]

hhorak:
Thanks, merging.

PR #187, opened 2017-12-18
Dockerfile: put variables inside curly braces

pkubatrh:
since OSBS had trouble expanding some during build

Also fixes version label for php56


omron93:
Only these labels aren't working well in OSBS or other should be changed too? (`summary`, `description`, `io.k8s.description`, `io.k8s.display-name`)

pkubatrh:
I ran into this issue with the `com.redhat.component` label as OSBS complained that it could not find `package (container) rh-php$PHP_VER_SHORT-docker` in the list of packages for the current rhscl tag.

It is obvious that this is an issue with expanding the variable, likely due to dash after `$PHP_VER_SHORT`? But I have not been able to reproduce this on any systems so I just changed any labels I thought might be also affected.

pkubatrh:
Looking at the built image all of the labels seem to be ok now.

Issue #189, opened 2017-12-21
/tmp/sessions permission denied


slbteam08:
I pulled the latest centos/php-71-centos7 and found my apps start showing permission denied errors on /tmp/session/:

`Warning: session_start(): open(/tmp/sessions/sess_eb242dea5a72c0892fa8758d86836ec0, O_RDWR) failed: Permission denied (13)
`

I looked that code and found that seems caused by a recent change which introduced the /usr/libexec/container-setup script. In this script /tmp/sessions owner is set to 1000:0. It used to be 1001:0.

Sorry that I am not familiar with steps to contribute code change. Could anyone help fixing the issue?

omron93:
@slbteam08 Thanks for reporting this problem. Contribution is welcomed - simply create a PR. One basic rule is that we try to keep supported versions similar as possible (to keep maintenance simpler) - https://github.com/sclorg/welcome/blob/master/contribution.md#active-versions. So please do changes in all version listed in `VERSIONS` in https://github.com/sclorg/s2i-php-container/blob/master/Makefile

@hhorak Was setting user `1000` for `/tmp/sessions` intentional?

hhorak:
> @hhorak Was setting user 1000 for /tmp/sessions intentional?

It was definitely not intentional O:-) Thanks both for spotting and analysis. PR for fixing this prepared: #191.

omron93:
@slbteam08 Should be fixed. If the problem still persist please reopen the issue.

slbteam08:
@omron93 I tested with the updated image and saw the owner changed, but still getting "permission denied" errors. Maybe "chmod a+rwx /tmp/sessions" is needed as well?

omron93:
Again. @slbteam08 This should be fixed. If the problem still persist please reopen the issue.

PR #193, opened 2018-01-05
Fix permissions for /tmp/sessions

slbteam08:
Try to fix issue #189.
`chmod -R a+rwx /tmp/sessions` was run before the code moved to /usr/libexec/container-setup.

### DIFF
--- a/5.6/root/usr/libexec/container-setup
+++ b/5.6/root/usr/libexec/container-setup
@@ -38,6 +38,7 @@ chmod -R a+rwx ${APP_ROOT}/etc
 chmod -R a+rwx ${HTTPD_VAR_RUN}
 chown -R 1001:0 ${APP_ROOT}
 mkdir /tmp/sessions
+chmod -R a+rwx /tmp/sessions
 chown -R 1001:0 /tmp/sessions
 chown -R 1001:0 ${HTTPD_DATA_PATH}
 chmod -R a+rwx ${PHP_SYSCONF_PATH}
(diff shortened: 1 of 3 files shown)


omron93:
[test] [test-openshift]

slbteam08:
Are there any ways I could check what's wrong? The https://coreservices-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhscl-images-php-rh/96/ is not accessible.

omron93:
There is no test for this issue and It would be great to have one. @slbteam08 Do you have any idea how to test this issue?

> Are there any ways I could check what's wrong? The https://coreservices-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhscl-images-php-rh/96/ is not accessible.

Currently it's not possible. It's on my TODO list, but I haven't managed it yet. Sorry.

slbteam08:
The problem is hit whenever a PHP session is started. But I do not know how to write a test case. Need some time to study.

omron93:
I don't see any other problem there. So merging. I hope it'll fix the problem.

> The problem is hit whenever a PHP session is started.

And what is a difference? Do you have any idea why the warning message isn't in our CI?

slbteam08:
When a PHP session is started, session file is created in /tmp/session by default. Seems that the test app does not start a session?

omron93:
> When a PHP session is started, session file is created in /tmp/session by default. Seems that the test app does not start a session?

@slbteam08 Could you have some time please to create s2i test app which starts session? And if you don't want to spent time to figure out how testing works I can update `test/run` to process this app there.

slbteam08:
OK, I will study existing code and try to create a test app.

PR #196, opened 2018-01-23
Add test app and test cases for /tmp/session permissions

slbteam08:
Follow up of PR #193. Add test cases for the issue #189.


omron93:
@slbteam08 Thanks for the tests. LGTM.

PR #197, opened 2018-01-23
WIP - Adding php 7.1 Nginx s2i

bobbydeveaux:
WORK IN PROGRESS- Comments most welcome!

It’s off the back of working with 5-6 clients all using PHP who want to use an official RHSCL image, but also need nginx. Some of them also need frontend npm/grunt stuff which Im looking at a nicer way to do that too. Not everyone is totally micro-service ready :/


hhorak:
@remicollet @notroj @uhliarik Guys, what do you think?

ElytschaOne:
If we move from Apache to a nginx-php-fpm env, it would be nice to achive max performance, so it would be better to work with an unix sockets instead of connecting nginx and php-fpm via IPv4

dakkusingh:
++

omron93:
@remicollet @notroj @uhliarik ?

notroj:
Thanks for submitting this @bobbydeveaux - interesting.

So what I've struggled with here before is not being sure how best to run multiple processes within the container - AIUI the approach here has a new services.sh which spawns php-fpm as a daemon and then execs nginx?  What happens if php-fpm dies, will docker notice?

phracek:
Closing this pull request. PHP-7.1 is no longer supported.
Feel free to reopen it or open the newer one against the supported version.

PR #198, opened 2018-01-27
[WIP] First try of conu test implementation

hhorak:
Based on https://github.com/sclorg/s2i-ruby-container/pull/167 with slight generalization. The point here is to see how much of the tests we can actually generalize, and put into conu library. It seems quite a lot.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 # Include common Makefile code.
 BASE_IMAGE_NAME = php
+TEST_IMAGE_NAME := $(BASE_IMAGE_NAME)-tests
+TEST_TARGET := test/test_s2i.py
 VERSIONS = 5.6 7.0 7.1
 OPENSHIFT_NAMESPACES = 5.5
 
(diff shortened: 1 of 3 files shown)


hhorak:
@tomastomecek, FYI, we might chat about this sometime.

Makefile
```diff
@@ -8,3 +10,10 @@ OPENSHIFT_NAMESPACES = 5.5
 .PHONY: $(shell test -f common/common.mk || echo >&2 'Please do "git submodule update --init" first.')
 
 include common/common.mk
+
+build-test-container:
+	docker build --network host --tag=$(TEST_IMAGE_NAME) -f ./test/Dockerfile.tests .
```
TomasTomecek: `Dockerfile.tests` seems to be missing

test/test_s2i.py
```diff
@@ -0,0 +1,86 @@
...
+def app(request):
+    i = S2IDockerImage(image_name)
+    app_name = os.path.basename(request.param)
+    app = i.extend(request.param, app_name)
+    yield app
+    pass
+    app.rmi()
+
+
+class GenericTestSuite:
```
TomasTomecek: I'm assuming you want this class to be in conu :)

test/test_s2i.py
```diff
@@ -0,0 +1,86 @@
...
+test_app_path = os.path.join(test_dir, "test-app")
+app_paths = [
+    test_app_path
+]
+
+
+backend = DockerBackend(logging_level=logging.DEBUG)
+
+
+@pytest.fixture(scope="module", params=app_paths)
```
TomasTomecek: and we could also start shipping some fine-crafted pytest fixtures in conu

Makefile
hhorak: Correct, will do.

test/test_s2i.py
hhorak: Yes, that was the intention, but I did not go far enough to propose it as a PR.

hhorak:
> otherwise lgtm, but I'm assuming you want to talk to me

Since there is no pressure to have this PR finished asap, I'd like rather to do the changes in `conu` first and then use the generic functions here, instead redefining the same thing here again. So yeah, let's chat before :)

phracek:
We have disabled conu in RHSCL CI testing. Closing this issue.

PR #216, opened 2018-06-01
Document npm purpose and no guarantee for version

hhorak:
SSIA


hhorak:
The same change reviewed in other images, and tests fail because of other reasons, text change should not cause any troubles. So, merging.

PR #217, opened 2018-06-05
Use released repos only

hhorak:
It is not necessary to use testing repo now anymore, all packages are in released repo already.


omron93:
Fedora CI isn't failing because of this PR. Important is that centos images is building well. So merging.

PR #220, opened 2018-07-09
imagestream definition for rhel based php 7.1 on ppc64le

jeffdyoung:
Adding this rhel7 php 7.1 imagestream that will support ppc64le.


hhorak:
LGTM, thanks.

Issue #230, opened 2018-10-15
Allow tuning php.ini's memory_limit


ibotty:
It would be great to be able to set `memory_limit`: https://github.com/sclorg/s2i-php-container/blob/master/7.1/root/opt/app-root/etc/php.ini.template#L389

I can provide a patch, but what's the preferred environment variable name?  I prefer `PHP_MEMORY_LIMIT`.

Cheers

omron93:
@remicollet ?

remicollet:
As 128M is really big, make sense

ibotty:
Well, for the one app I am currently deploying it's too small :(.

omron93:
@ibotty :-D Anyway, PR is really welcomed. `PHP_MEMORY_LIMIT` sound fine. Thanks.

PR #229, opened 2018-10-05
Fix Fedora based image - switch back to 'prefork'

omron93:
Fedora uses by default 'event' MPM module, so switch to 'prefork'
to provide same user experience as with RHEL/CentOS image.

For more information see https://bugzilla.redhat.com/show_bug.cgi?id=1586332 .

@pkubatrh @notroj @remicollet @dustymabe Please review. Any ideas are welcomed!

PR #231, opened 2018-10-17
Make PHP.ini's memory_limit configurable

ibotty:


omron93:
Resolves: #230

[test][test-openshift]

omron93:
@ibotty Thanks for this improvement.

Fedora CI fails because of unmerged #229 and rhel7 openshift CI because of outdated common/. This change was discussed in #230, so merging.

PR #235, opened 2019-01-23
PHP72

madrisan:
This patch adds the PHP 7.2 version, now available in Red Hat Software Collection.


hhorak:
Thanks for the contribution.

One thing I spotted: The `.*image-id*` files are actually not supposed to be part of the repo, so those can be either removed from the PR, or it would be probably better to add them to the `.gitignore`, similar as it is done for postgresql image: https://github.com/sclorg/postgresql-container/blob/master/.gitignore

hhorak:
[test]

hhorak:
[test-openshift]

pkubatrh:
Thanks for the PR!

php 7.2 is in the process of being added via https://github.com/sclorg/s2i-php-container/pull/239 so I will be closing your PR in favour of https://github.com/sclorg/s2i-php-container/pull/239.

PR #246, opened 2019-05-26
fix: respect PHP_MEMORY_LIMIT for composer

PixelJonas:
After setting PHP_MEMORY_LIMIT to `-1` composer still exited with an OOM error.
Setting the parameter directly through php fixed the issue


PixelJonas:
any chance this is getting merged? I'm currently monkey-patching the `assemble` file :(

PixelJonas:
This is still an open issue for me? can anyone verify this patch?

PixelJonas:
any chance this will ever get merged?

phracek:
[test-all]

phracek:
@PixelJonas Thanks for the PR!. Any documentation about this change? I am not familiar with php ;)

phracek:
@PixelJonas Can you please include it into the other versions, 7.4?

phracek:
[test-all]

phracek:
@remicollet Does it make sense for you this PR?

@PixelJonas Can you please add this change also for (https://github.com/sclorg/s2i-php-container/blob/master/8.0/s2i/bin/assemble)

[test-all]

phracek (changes requested):
Please add this also into 8.0 version.

PixelJonas:
I added the patch to 8.0

remicollet:
I can understand the need to set memory limit for cli command (e.g. composer)
but I don't think it make sense to use the same value for web process (usually quite low) and for cli process (usually quite high)

It could be better to use 2 configuration options
* `PHP_MEMORY_LIMIT` for web process, set using php_value in httpd.conf (mod_php) or www.conf (php-fpm)
* `PHP_MEMORY_LIMIT_CLI` set for other process in php.ini

BTW, for now  `PHP_MEMORY_LIMIT` is set is php.ini, so should be used by `composer`, which mean php.ini is NOT properly populated (after composer step ?)

phracek:
@PixelJonas Can you please fix it as @remicollet propose two solutions?

PixelJonas:
That heavily depends on the timeframe you are expecting this to be done. Right now I don't see the bandwidth to do that, as this looks like something which needs to be tested and I don't have a scenario this is actually being used in (this PR is from 2019).

PR #260, opened 2019-10-07
all

samyunodos:


phracek:
@samyunodos Sorry, but your PR a bit strange. Please follow up guideline https://github.com/sclorg/welcome/blob/master/contribution.md. I would prefer to rebase it and squash commits into one.

PR #269, opened 2019-10-24
Add support for internal Nexus registry

phracek:
Add parameter `ct_build_s2i_npm_variables`
into `run_s2i_build` function.
It checks if internal NPM_REGISTRY is set and
CA file is present. Else use public NPM registry.


phracek:
reference https://github.com/sclorg/s2i-php-container/pull/269

pkubatrh:
@phracek missing submodule update, so the pr did not likely have the changes from container-common-scripts, along with a bugfix from https://github.com/sclorg/container-common-scripts/pull/125

PR #270, opened 2019-10-28
all

samyunodos:


pkubatrh:
Hi @samyunodos 

While sclorg repositories are open to everyone for contribution, your recent contributions do not look like opened on purpose. I think this is caused by some automation on your side and you really did not want to create them, seeing as they have string translations without English alternatives, so please fix your scripts.

If you continue opening up PRs against sclorg repositories that provide no value I will take them as spam and will be forced to block your account from accessing the repositories.

Thanks for understanding.

PR #271, opened 2019-11-04
Expose sendmail_path as environment variable.

fengtan:
Is it possible to expose the `sendmail_path` PHP setting as an environment variable ?
It would allow applications running on this image to send emails to arbitrary mail services like [Mailhog](https://github.com/mailhog/MailHog).


phracek:
[test-all]

phracek:
@fengtan Can you please update PR against master?

PR #292, opened 2020-06-17
Test for check the latest version in imagestreams

phracek:
This commit adds check for the latest version in imagestreams.

If the latest version is not present in imagestreams, then the test fails.


phracek:
Closing this PR. There are some troubles with container-common-scripts. I will create a new one.

PR #294, opened 2020-06-22
Run all tests by default

phracek:
This commit runs all tests by default and report only FAILUREs.


hhorak:
I see no issue here, and passing test is also a good enough feedback. LGTM.

PR #297, opened 2020-07-21
container tests finish with TESTSUITE_RESULT

phracek:
In case some of the test fails, test suite finishes
with TESTSUITE_RESULT value.

check_result function is used for setting TESTSUITE_RESULT value.


hhorak:
LGTM

PR #309, opened 2020-09-25
Document usage with a Dockerfile

hhorak:
This adds the following:

- Dockerfile examples for a variant that uses s2i scripts and a variant without s2i scripts
- Documentation enhancement that describes how the package can be used using a Dockerfile a test for the example Dockerfiles

This is a similar pull-request as sclorg/httpd-container#99 and https://github.com/sclorg/s2i-nodejs-container/pull/247.

To see the formatted README.md, see https://github.com/hhorak/s2i-php-container/blob/doc-dockerfile/7.3/README.md and it is also visible on the catalog, which will be the most important view: https://catalog.redhat.com/software/containers/detail/5d4008a7dd19c75498b19083


hhorak:
@phracek @pkubatrh @fila43 FYI, it follows pretty closely wording in https://github.com/sclorg/s2i-nodejs-container/pull/247 and tests pass, so hopefully should be good, but some recheck might be useful.

PR #313, opened 2020-11-03
Add COMPOSER_INSTALLER_OPTIONS to specify version of Composer etc.

andreasottosson:
Composer install defaults to version 2. Adding support for using version 1.

Please add this to other versions, eg 7.4 as you see fit.

### DIFF
--- a/7.3/README.md
+++ b/7.3/README.md
@@ -244,6 +244,8 @@ yourself:
       * Adds a custom composer repository mirror URL to composer configuration. Note: This only affects packages listed in composer.json.
     * **COMPOSER_INSTALLER**
       * Overrides the default URL for downloading Composer of https://getcomposer.org/installer. Useful in disconnected environments.
+    * **COMPOSER_VERSION**
+      * Overrides the default Composer version to download. See https://getcomposer.org/download/ for releases. Useful if you need version 1 or specific version.
     * **COMPOSER_ARGS**
       * Adds extra arguments to the `composer install` command line (for example `--no-dev`).
 
--- a/7.3/README.md
+++ b/7.3/README.md
@@ -245,7 +245,7 @@ yourself:
     * **COMPOSER_INSTALLER**
       * Overrides the default URL for downloading Composer of https://getcomposer.org/installer. Useful in disconnected environments.
     * **COMPOSER_VERSION**
-      * Overrides the default Composer version to download. See https://getcomposer.org/download/ for releases. Useful if you need version 1 or specific version.
+      * Overrides the default Composer version to download. Set to --1 for latest Composer v1, --2 for latest Composer v2. Defaults to v2. For specific versions see https://getcomposer.org/download/.
     * **COMPOSER_ARGS**
       * Adds extra arguments to the `composer install` command line (for example `--no-dev`).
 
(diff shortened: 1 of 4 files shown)


rhscl-automation-poll:
Can one of the admins verify this patch?

rhscl-automation-poll:
Can one of the admins verify this patch?

rhscl-automation-poll:
Can one of the admins verify this patch?

rhscl-automation-poll:
Can one of the admins verify this patch?

blazmanx34:
It will be better to make this :

```
  if [ -z "$COMPOSER_VERSION" ]; then
    export COMPOSER_VERSION_ARG=""
  elif [ $COMPOSER_VERSION = "--1" ]; then
    export COMPOSER_VERSION_ARG="--1"
  elif [ $COMPOSER_VERSION = "--2" ]; then
    export COMPOSER_VERSION_ARG="--1"
  else
    export COMPOSER_VERSION_ARG="--version=$COMPOSER_VERSION"
  fi

  php $TEMPFILE $COMPOSER_VERSION_ARG
```

By this way, it allows to install the last version 1, the last version 2 or a specific version.
(The old behavior is preserved : install the latest version of composer, the last version 1 or the last version 2)

andreasottosson:
Awesome! I was looking for an argument that would fetch the latest but I didn't find it, looks great! I will update the PR.

Do you think we should have the user set COMPOSER_VERSION="--1" or just COMPOSER_VERSION="1"?

andreasottosson:
PR updated, I also added the changes to 7.4

meanderix:
@andreasottosson @blazmanx34 Thanks for this PR! Looking forward to this being merged.

tzutzurel:
Hi,
you can anticipate v3 :) 

if [ -z "$COMPOSER_VERSION" ]; then
  export COMPOSER_VERSION_ARG=""
elif [[ $COMPOSER_VERSION = --* ]] ; then
  export COMPOSER_VERSION_ARG=$COMPOSER_VERSION
else
  export COMPOSER_VERSION_ARG="--version=$COMPOSER_VERSION"
fi

andreasottosson:
Nice @tzutzurel ! :)

Is it really necessary though? It took about 8 years for composer to go from v1 to v2. I feel like the documentation part gets more complex if we say something like, "You can set --* where * is the version of Composer to use..."

What do you guys think? I can easily update the PR with this solution if that is what we want.

tzutzurel:
It was a joke ....
But, the last solution allows to get full control on the parameters ....

COMPOSER_VERSION="--key1=value1 --key2=value2 ..."

Maybe a better name like COMPOSER_INSTALL_PARAMS ?

And you can avoir errors like @blazmanx34 

  elif [ $COMPOSER_VERSION = "--2" ]; then
    export COMPOSER_VERSION_ARG="--1"

:)

andreasottosson:
Are we over-engineering this? Could we not simply do this

` php $TEMPFILE $COMPOSER_INSTALLER_OPTIONS`

And in the docs we point to https://getcomposer.org/download/, the options could be --filename, --version, --install-dir and of course just --1 for v1 and --2 for v2. Leave blank for v2 which is the default. It also works with multiple options "--1 --install-dir=/tmp", specific version "--version=1.10.17" and so on.

andreasottosson:
PR updated again, this should be good to merge.

phracek:
@andreasottosson We have already merged this PR  https://github.com/sclorg/s2i-php-container/pull/339. If it is fine for you then close this PR. Thank

PR #322, opened 2020-12-09
Update s2i usage script to use OpenShift use case

hhorak:


pkubatrh:
Looking at test results, only Dockerfile test case failed because of quay.io, so LGTM.

pkubatrh:
Thanks for the PR!

PR #332, opened 2021-05-31
Deprecate PHP-72 image for RHEL8 which reached EOL in May 2021

phracek:


phracek:
[test][test-openshift-4]

hhorak:
LGTM

PR #333, opened 2021-07-15
Add support for Testing Farm CI

phracek:
This pull request adds support for testing s2i-php-container in Testing Farm

Supported scenarios are:

Fedora
CentOS7
CentOS7 with OpenShift

How does it work?
As soon as a pull request is created, when the owner or member writes the comment [test], then Testing Farm is scheduled.


phracek:
We can use https://github.com/sclorg/.github repository with actions.

PR #334, opened 2021-07-19
Fixes ini scan usage

paulbadcock:
Per the documentation

```text
PHP_INI_SCAN_DIR
Path to scan for additional ini configuration files
```

If you set this value on a running container it will no longer include ${PHP_SYSCONF_PATH}/php.d. This will break the container as normally this directory is populated with the pre-baked in PHP modules.

This change includes the PHP_SYSCONF_PATH so you can include an additional directory in the app code base to scan.

If rejected documentation should be updated to clearly indicate you still need to include the base directory for your container. But this causes a further issue as you have to maintain the exact pathing to the php version compiled into the image.

### DIFF
--- a/7.3/s2i/bin/run
+++ b/7.3/s2i/bin/run
@@ -29,7 +29,12 @@ export OPCACHE_REVALIDATE_FREQ=${OPCACHE_REVALIDATE_FREQ:-2}
 export OPCACHE_MAX_FILES=${OPCACHE_MAX_FILES:-4000}
 
 export PHPRC=${PHPRC:-${PHP_SYSCONF_PATH}/php.ini}
-export PHP_INI_SCAN_DIR=${PHP_INI_SCAN_DIR:-${PHP_SYSCONF_PATH}/php.d}
+
+if [ -z "${PHP_INI_SCAN_DIR:-}" ]; then
+   export PHP_INI_SCAN_DIR=${PHP_SYSCONF_PATH}/php.d
+else
+   export PHP_INI_SCAN_DIR=${PHP_INI_SCAN_DIR}:${PHP_SYSCONF_PATH}/php.d
+fi
 
 envsubst < /opt/app-root/etc/php.ini.template > ${PHP_SYSCONF_PATH}/php.ini
 envsubst < /opt/app-root/etc/php.d/10-opcache.ini.template > ${PHP_SYSCONF_PATH}/php.d/10-opcache.ini
--- a/7.4/s2i/bin/run
+++ b/7.4/s2i/bin/run
@@ -29,7 +29,12 @@ export OPCACHE_REVALIDATE_FREQ=${OPCACHE_REVALIDATE_FREQ:-2}
 export OPCACHE_MAX_FILES=${OPCACHE_MAX_FILES:-4000}
 
 export PHPRC=${PHPRC:-${PHP_SYSCONF_PATH}/php.ini}
-export PHP_INI_SCAN_DIR=${PHP_INI_SCAN_DIR:-${PHP_SYSCONF_PATH}/php.d}
+
+if [ -z "${PHP_INI_SCAN_DIR:-}" ]; then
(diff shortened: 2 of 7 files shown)


phracek:
@paulbadcock Thanks for the PR!. But only php-73 and php-74 are supported. Please rebase so only supported versions are updated.

Is there any way how to test PHP_INI_SCAN_DIR  by our test suite https://github.com/sclorg/s2i-php-container/blob/master/test/run ?

paulbadcock:
> @paulbadcock Thanks for the PR!. But only php-73 and php-74 are supported. Please rebase so only supported versions are updated.
> 
> Is there any way how to test PHP_INI_SCAN_DIR by our test suite https://github.com/sclorg/s2i-php-container/blob/master/test/run ?

Reverted the changes so only the supported versions are changed.

It looks like the testing suite has something for `ct_check_envs_set` but nothing is in this repo to validate environment variables. Is it used on other s2i projects or am I missing where the validation of environments are tested in this repo?

phracek:
@paulbadcock Can you please update also a documentation with PHP_SYSCONF_PATH. I guess the user should know it.
This will  be last change from my side.

Let's check the tests

[test]

phracek:
@paulbadcock Can you please update also the documentation with PHP_SYSCONF_PATH. I guess the user should know it.
This will be the last change from my side.

Let's check the tests

[test]

phracek:
@paulbadcock Can you please update also the documentation with PHP_SYSCONF_PATH. I guess the user should know it.
This will be the last change from my side.

Let's check the tests

[test]

phracek:
[test]

phracek:
[test]

paulbadcock:
Fixed the documentation to the correct directories.

7.3/s2i/bin/run
```diff
@@ -29,7 +29,12 @@ export OPCACHE_REVALIDATE_FREQ=${OPCACHE_REVALIDATE_FREQ:-2}
 export OPCACHE_MAX_FILES=${OPCACHE_MAX_FILES:-4000}
 
 export PHPRC=${PHPRC:-${PHP_SYSCONF_PATH}/php.ini}
-export PHP_INI_SCAN_DIR=${PHP_INI_SCAN_DIR:-${PHP_SYSCONF_PATH}/php.d}
+
+if [ -z "${PHP_INI_SCAN_DIR:-}" ]; then
```
phracek, line 33:
```suggestion
if [ -z "${PHP_INI_SCAN_DIR}" ]; then
```

This should be enough, for testing if `PHP_INI_SCAN_DIR` variable is empty.

7.4/s2i/bin/run
```diff
@@ -29,7 +29,12 @@ export OPCACHE_REVALIDATE_FREQ=${OPCACHE_REVALIDATE_FREQ:-2}
 export OPCACHE_MAX_FILES=${OPCACHE_MAX_FILES:-4000}
 
 export PHPRC=${PHPRC:-${PHP_SYSCONF_PATH}/php.ini}
-export PHP_INI_SCAN_DIR=${PHP_INI_SCAN_DIR:-${PHP_SYSCONF_PATH}/php.d}
+
+if [ -z "${PHP_INI_SCAN_DIR:-}" ]; then
```
phracek, line 33:
```suggestion
if [ -z "${PHP_INI_SCAN_DIR}" ]; then
```

This should be enough, for testing if `PHP_INI_SCAN_DIR` variable is empty.

phracek (changes requested):
I guess, checking if variable is empty it is enough to write`if [ -z "${VAR} ]]; then`

https://www.cyberciti.biz/faq/unix-linux-bash-script-check-if-variable-is-empty/

phracek:
[test-openshift][test-openshift-4]

phracek:
@paulbadcock Can you please look at solved my comments?

PR #348, opened 2022-01-18
Add support CentOS Stream 9 to GitHub Actions

phracek:
This pull request adds support CentOS Stream 9 to GitHub Actions.

s2i-php-container is now tested also in CentOS Stream 9.
And the php-8.0 image is also pushed to quay.io repository.


zmiklank (approved):
Thank you for splitting the original PR. This LGTM.

PR #349, opened 2022-01-18
Fix tmt_url

phracek:
tmt_url should not end fix '/' and tests are present in artifacts.dev.... url.


phracek:
This can be merged. Affects only GitHub Actions.

PR #347, opened 2022-01-13
Support for CentOS Stream 9

phracek:
This pull request add support for testing, building and pushing s2i-php-container into Quay.io.

The CentOS Stream 9 image is pushed to quay.io/sclorg namespace.

@hhorak @pkubatrh @zmiklank WDYT?

PR #354, opened 2022-03-23
use composite build-and-push-action

zmiklank:
- rebuild images every wednesday
- enable action only for sclorg repositories

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -1,79 +1,40 @@
-name: Build and push CentOS7 images to Quay.io registry
+name: Build and push images to Quay.io registry
 on:
   push:
     branches:
       - master
+  schedule:
+      - cron: '0 1 * * 3'
 
 jobs:
   build-and-push:
+    if: github.repository_owner == 'sclorg'
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        version: [7.3, 7.4, 8.0]
+        dockerfile_path: ["7.3", "7.4"]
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -1,79 +1,39 @@
...
+      - cron: '0 1 * * 3'
 
 jobs:
   build-and-push:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        version: [7.3, 7.4, 8.0]
+        dockerfile_path: ["7.3", "7.4", "8.0"]
```
phracek, line 15: Version '8.0' is not part of the repository yet ;)

phracek (changes requested):
Just version '8.0' is not part of upstream. It will be as soon as https://github.com/sclorg/s2i-php-container/pull/347 is merged.

.github/workflows/build-and-push.yml
zmiklank: Interesting, because it was in the build-and-push even before my proposal. should I remove it?

phracek: @zmiklank Please remove version 8.0. It is not part of `master` branch yet ;)

phracek (changes requested):

.github/workflows/build-and-push.yml
zmiklank: Done, sorry for the delay.

phracek (approved):
LGTM

PR #365, opened 2022-06-03
Allow building and testing RHEL8 PHP-80

phracek:

### DIFF
--- a/8.0/s2i/bin/run
+++ b/8.0/s2i/bin/run
@@ -57,11 +57,16 @@ fi
 
 
 if [ "x$PLATFORM" == "xel9" ] || [ "x$PLATFORM" == "xfedora" ]; then
-  if [ -z "${PHP_FPM_LOG_PATH:-}" ]; then
+  if [ -n "${PHP_FPM_RUN_DIR:-}" ]; then
     /bin/ln -s /dev/stderr ${PHP_FPM_LOG_PATH}/error.log
+    mkdir -p ${PHP_FPM_RUN_DIR}
+    chmod -R a+rwx ${PHP_FPM_RUN_DIR}
+    chown -R 1001:0 ${PHP_FPM_RUN_DIR}
+    mkdir -p ${PHP_FPM_LOG_PATH}
+    chmod -R a+rwx ${PHP_FPM_LOG_PATH}
+    chown -R 1001:0 ${PHP_FPM_LOG_PATH}
   fi
-  chmod -R a+rwx ${PHP_FPM_LOG_PATH}
-  chown -R 1001:0 ${PHP_FPM_LOG_PATH}
+
 fi
 
 # pre-start files
--- a/8.0/s2i/bin/run
+++ b/8.0/s2i/bin/run
@@ -54,12 +54,15 @@ else
   echo "-> Cgroups memory limit is set, using HTTPD_MAX_REQUEST_WORKERS=${HTTPD_MAX_REQUEST_WORKERS}"
 fi
 
-if [ ! -v PHP_FPM_LOG_PATH ]; then
-  /bin/ln -s /dev/stderr ${PHP_FPM_LOG_PATH}/error.log
-fi
 
-chmod -R a+rwx ${PHP_FPM_LOG_PATH}
-chown -R 1001:0 ${PHP_FPM_LOG_PATH}
+
+if [ "x$PLATFORM" == "xel9" ] || [ "x$PLATFORM" == "xfedora" ]; then
+  if [ -z "${PHP_FPM_LOG_PATH:-}" ]; then
+    /bin/ln -s /dev/stderr ${PHP_FPM_LOG_PATH}/error.log
+  fi
+  chmod -R a+rwx ${PHP_FPM_LOG_PATH}
+  chown -R 1001:0 ${PHP_FPM_LOG_PATH}
+fi
 
 # pre-start files
 process_extending_files ${APP_DATA}/php-pre-start/ ${PHP_CONTAINER_SCRIPTS_PATH}/pre-start/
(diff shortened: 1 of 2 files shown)


phracek:
[test-all]

phracek:
[test-all]

phracek:
@remicollet @hhorak @pkubatrh Can anyone please review it?

8.0/s2i/bin/run
```diff
@@ -54,12 +54,15 @@ else
...
 
-if [ ! -v PHP_FPM_LOG_PATH ]; then
-  /bin/ln -s /dev/stderr ${PHP_FPM_LOG_PATH}/error.log
-fi
 
-chmod -R a+rwx ${PHP_FPM_LOG_PATH}
-chown -R 1001:0 ${PHP_FPM_LOG_PATH}
+
+if [ "x$PLATFORM" == "xel9" ] || [ "x$PLATFORM" == "xfedora" ]; then
+  if [ ! -v PHP_FPM_LOG_PATH ]; then
```
remicollet, line 60: What is expected here ?
* don't know about -v test
* should probably be ${PHP_FPM_LOG_PATH}

phracek: See a check for the presence of variable here: https://www.cyberciti.biz/faq/see-check-if-bash-variable-defined-in-script-on-linux-unix-macos/ . Chapter `Checking if a variable is set in Bash or not`

phracek: The -v VAR returns true if the shell variable VAR is set:
```
[[ -v ${var} ]] && echo "\${var} not defined" 
[[ -v ${var} ]] && echo "\${var} not defined" || echo "\${var} is defined"
[ -v ${var} ] && echo "\${var} not defined" || echo "\${var} is defined"
```

remicollet: I mean -v is not in "test" man page ;)   (and I have never used it, despite ~30y of shell ;)

BTW in all cases, `[ ! -v PHP_FPM_LOG_PATH ]` is wrong, missing **$**

pkubatrh: I would argue to use `-z` or `-n` here. Afaik `-v` only checks if the variable is set, but I do not think we want to continue in case it is empty.

pkubatrh: I also find the use of `[ ! -v ]` here weird - we check if the variable is not set and then we use it for a symlink path? Am I missing something crucial?

phracek: I have resolved it by commit https://github.com/sclorg/s2i-php-container/pull/365/commits/0b18c2a56d66d034ef8edfdc4c45bfc7b502de68

phracek:
[test-all]

phracek:
@remicollet @pkubatrh Can you please take a look at it one more time?

8.0/s2i/bin/run
```diff
@@ -54,12 +54,15 @@ else
...
-  /bin/ln -s /dev/stderr ${PHP_FPM_LOG_PATH}/error.log
-fi
 
-chmod -R a+rwx ${PHP_FPM_LOG_PATH}
-chown -R 1001:0 ${PHP_FPM_LOG_PATH}
+
+if [ "x$PLATFORM" == "xel9" ] || [ "x$PLATFORM" == "xfedora" ]; then
+  if [ -z "${PHP_FPM_LOG_PATH:-}" ]; then
+    /bin/ln -s /dev/stderr ${PHP_FPM_LOG_PATH}/error.log
+  fi
```
remicollet, line 62: I still don't understand this test...
if empty, use it.... sounds strange

8.0/s2i/bin/run
```diff
@@ -54,12 +54,15 @@ else
...
+  chmod -R a+rwx ${PHP_FPM_LOG_PATH}
+  chown -R 1001:0 ${PHP_FPM_LOG_PATH}
```
remicollet, line 64: Shouldn't this command be in the above test about PHP_FPM_LOG_PATH being set

8.0/s2i/bin/run
phracek: Sorry, wrong bash condition ;) You are right ;) Variable is different. It should be https://github.com/sclorg/s2i-php-container/blob/master/8.0/Dockerfile.rhel9#L60.

I have to review a bit deeply.

phracek: See https://github.com/sclorg/s2i-php-container/search?q=PHP_FPM_RUN_DIR

8.0/s2i/bin/run
phracek: You are right, it should be fixed commit https://github.com/sclorg/s2i-php-container/commit/122b25fe24807d4cc7aec24fdc62a262e78aa396

8.0/s2i/bin/run
phracek: IT should be fixed by commit https://github.com/sclorg/s2i-php-container/commit/122b25fe24807d4cc7aec24fdc62a262e78aa396

phracek:
[test-all]

phracek:
[test-all]

phracek:
@pkubatrh @remicollet Can you please look at it one more time?

[test-all]

phracek:
[test-all]

phracek:
[test-all]

phracek:
RHEL9 unsubscribed host hit a problem with tzdata:

```bash
This system is not registered with an entitlement server. You can use subscription-manager to register.

Last metadata expiration check: 0:00:06 ago on Wed Sep 14 11:01:54 2022.
Installed package tzdata-2022a-1.el9_0.noarch not available.
Error: No packages marked for reinstall.
```

phracek:
[test-all]

phracek:
@pkubatrh @remicollet PTAL

remicollet:
LGTM

PR #368, opened 2022-06-09
RPM uses a wrong file in the config, fix it in the container

hhorak:
Related: https://bugzilla.redhat.com/show_bug.cgi?id=2092356


hhorak:
[test-all]

PR #376, opened 2022-07-26
On httpd startup solve AH02282: No slotmem

olstjos:


olstjos:
> test @phracek 

yes?

phracek (changes requested):
Thanks for this pull request. But what about to allow it like e.g. in this case. 
https://github.com/sclorg/s2i-php-container/pull/406

PR #382, opened 2022-10-04
Use test-openshift.yaml from common dir.

phracek:
Do not use double symlinks


phracek:
This fix did not break any CI. But for sure let's test it.

[test-all]

PR #388, opened 2022-11-15
Add a test in Dockerfile directly to check php version

hhorak:
Otherwise when a wrong stream is installed, it's only found by scl test, which is too late in the process and might even go through uncought, especially when the scl test is removed for non-SCL distros at some point.


hhorak:
[test]

hhorak:
[test]

phracek:
The RHEL9 test is failing on:
```bash
Last metadata expiration check: 0:00:11 ago on Mon Jan 16 16:32:49 2023.
Installed package tzdata-2022f-1.el9_1.noarch not available.
Error: No packages marked for reinstall.
```

RHEL7 failure is expected. The problem with RHEL7 compose should be solved soon.

phracek:
Fedora failures are caused by Fedora 35 EOL.

phracek:
LGTM. Let's get merge it Fedora tests are fixed by https://github.com/sclorg/s2i-php-container/pull/390

PR #389, opened 2023-01-02
Fixes #387: Add the latest container-common-scripts

### LINKED ISSUE #387: Executing `test_scl_usage` when tty exists waits on 'cat' forever
From bash verbose when this test is executed:

```bash
[INFO] Testing the image SCL enable...

++ docker run --rm php-image /bin/bash -c ''
+ out=
+ echo ''
+ grep -q ''
+++ cat
```
and there it stops. When there is no tty cat exists immediately and test continues as expected.

phracek:
Closes #387


phracek:
[test-all]

Issue #374, opened 2022-07-12
PHP 8.1 build - It's the latest/greatest/fastest and we're anxious to start using it.


olstjos:
Locally I've tested my code using wodby on PHP 8.1 and 8.0, having a bit of trouble getting the s2i working in openshift/red hat for PHP 8.0, with that said, I'd much rather spend all this time getting PHP 8.1 going, any chance that PHP 8.1 builds will be ready soon?  There's additional performance improvements and reduced memory footprint available by using the latest and greatest PHP.

https://kinsta.com/blog/php-benchmarks/

pkubatrh:
Hi there @olstjos,
seems like the first module builds for PHP 8.1 are hitting Centos Stream 9 so we will be able to start working on the c9s version of the image quite soon. So if you are okay with using Centos Stream images, be on the lookout for the new image.

joejoseph00:
Hello @pkubatrh  I've got PHP 8.2.1 up and running with Ubuntu and it's blazingly fast, takes even less memory than PHP 8.1.

We need to get at least PHP 8.1 going here with the s2i-php-container because we're trying to run hundreds of pods on razor thin resources and we need to squeeze every single byte of ram and every cpu cycle to the maximum.  Our cloud billing is huge, our apps need to be faster and we need to get PHP 8.1 going asap.  PHP 8.0 will be end of life this year so we need to at least get started on PHP 8.1.

We're not running a bank with custom software, we are running cutting edge publicly available open source software and we need to stay current with security advisory releases and be ready for upgrades and everything the world is throwing at us.  Every extra bit of performance is improved availability and allows us to deal better with bots and concurrency and keep latency low.

pkubatrh:
Hi @joejoseph00 
PHP 8.1 has been available on RHEL9 for a good while, feel free to give it a try: https://catalog.redhat.com/software/containers/ubi9/php-81/62e8e662f6d3d47e19779b01

joejoseph00:
Hi @pkubatrh  thanks so much for looking into this.  
PHP 8.0 will become EOL on november 26th 2023, and we often start developing on beta releases that are scheduled to be released a week or two AFTER the EOL date.  As developers we need to be as far ahead of EOL as possible so that we can achieve our deliverables on time.  On november 26th 2023 we will lose access to upgrades of older version dependencies that we need to keep our application secure and long before this EOL date we also lose access to new functionality of the open source dependencies we rely on unless we develop using other platforms outside of the RHEL space.

[...]

PR #393, opened 2023-01-27
Add 8.1 versions for Fedora and RHEL9

pkubatrh:
Resolves: #374


pkubatrh:
[test-all]

pkubatrh:
[test-all]

pkubatrh:
[test-all]

pkubatrh:
[test-all]

pkubatrh:
[test-all]

pkubatrh:
[test-all]

pkubatrh:
[test-all]

phracek:
[test-all]

phracek:
Let's try it once more time. The subscription account should be now fixed.

[test-all]

8.1/Dockerfile.c9s
```diff
@@ -0,0 +1,85 @@
...
+
+EXPOSE 8080
+EXPOSE 8443
+
+# Description
+# This image provides an Apache 2.4 + PHP 7.4 environment for running PHP applications.
+# Exposed ports:
+# * 8080 - alternative port for http
+
+ENV PHP_VERSION=8.0 \
```
phracek, line 14: Wrong version. It should be `8.1`

8.1/Dockerfile.rhel8
```diff
@@ -0,0 +1,81 @@
...
+
+EXPOSE 8080
+EXPOSE 8443
+
+# Description
+# This image provides an Apache 2.4 + PHP 8.0 environment for running PHP applications.
+# Exposed ports:
+# * 8080 - alternative port for http
+
+ENV PHP_VERSION=8.0 \
```
phracek, line 14: This should be as well version `8.1` and not `8.0`

8.1/content_sets.yml
```diff
@@ -0,0 +1,10 @@
+# This is a file defining which content sets are needed to update content in
```
phracek, line 1: It this file really needed?

phracek (changes requested):
Some issues, which I have hit during the review.
Why do we have `8.1/.exclude-c9s` file and `8.1/Dockerfile.c9s` In case it is UBI based, then the `Dockerfile` and `exclude` file is not needed. This is a duplicity of `Dockerfile.rhel9` which can be downloaded as UBI image.

8.1/Dockerfile.c9s
pkubatrh: As 8.1 is already released we will not be building c9s version at all. The Dockerfile is there for the sole reason of not losing history for it.

8.1/Dockerfile.rhel8
pkubatrh: Similarly to c9s, we are not building 8.1 for RHEL8.

8.1/content_sets.yml
pkubatrh: Good point, will remove

8.1/Dockerfile.rhel8
phracek: Then we should either remove it or correct it. It is misleading.

8.1/Dockerfile.c9s
phracek: It is misleading to have `8.1/Dockerfile.c9s` and inside wrong PHP version.

pkubatrh: I do not see how it could be misleading to have old versions in unmaintained files (its not like we test them), but since it is a simple change to make I added both in a recent commit.

pkubatrh:
[test-all]

olstjos:
We're needing this still, what is remaining to do before this can pass tests and get merged?

olstjos:
Ya we're using OpenShift 
http://artifacts.osci.redhat.com/testing-farm/f91ba16d-8209-42bd-90af-696344d26a21/  this artifact is not visible to the public, I think it's internal to Red Hat.

8.1/Dockerfile.fedora
```diff
@@ -0,0 +1,80 @@
+FROM quay.io/fedora/s2i-base:36
```
phracek, line 1: Only nitpick. Can we used s2i-base:37? 
```suggestion
FROM quay.io/fedora/s2i-base:37
```

phracek: F36 is going to be EOL soon.

8.1/Dockerfile.fedora
```diff
@@ -0,0 +1,80 @@
...
+LABEL summary="$SUMMARY" \
+      description="$DESCRIPTION" \
+      io.k8s.description="$DESCRIPTION" \
+      io.k8s.display-name="Apache 2.4 with PHP $PHP_VERSION" \
+      io.openshift.expose-services="8080:http" \
+      io.openshift.tags="builder,php" \
+      name="fedora/$NAME-$PHP_SHORT_VER" \
+      com.redhat.component="$NAME" \
+      version="$VERSION" \
+      usage="s2i build https://github.com/sclorg/s2i-php-container.git --context-dir=/$PHP_VERSION/test/test-app quay.io/fedora/$NAME-$PHP_SHORT_VER sample-server" \
```
phracek, line 36: quay.io/fedora/php-81 is already created. Waiting on robot account. https://quay.io/repository/fedora/php-81?tab=settings

phracek (commented):
Only some nitpicks.

LGTM. Thanks. Let's wait for the tests.

pkubatrh:
[test-all]

phracek (approved):
LGTM. Thanks for adding new PHP version.

Let's wait whether container tests are passing. If so, then let's get merge it.

olstjos:
@phracek @pkubatrh I'm testing this again and got the same error I got when another colleague tested the RH9 , I'm wondering if something must have changed with regards to the docroot of the project.  AFAIK it used to be **/opt/app-root/src**
all our previous sclorg related builds are expecting this path as the docroot?  Which set of release notes should we look at for upgrading?

olstjos:
Hmm, which Dockerfile **from** value should be used?

previously our 8.0 was using:  

`FROM registry.access.redhat.com/ubi8/php-80`

olstjos:
Ok so I've tested this 8.1, and have found the following:

Our secret environment variables are showing up for the pod user but the web server instance does not see our environment variables.

This crashes our application because our server cannot connect to the database.

So somewhere in the 8.1 recipe, the web server user is no longer the same user that gets the openshift secrets.

PR #394, opened 2023-01-27
update github actions for 8.1

pkubatrh:
Needed for #393 

@phracek @zmiklank PTAL

### DIFF
--- a/.github/workflows/openshift-tests.yml
+++ b/.github/workflows/openshift-tests.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [ "7.3", "7.4", "8.0" ]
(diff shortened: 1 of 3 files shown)


.github/workflows/openshift-tests.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [ "7.3", "7.4", "8.0" ]
+        version: [ "7.3", "7.4", "8.0" , "8.1" ]
```
zmiklank, line 13: We do not test Fedora images on OpenShift. I think this should not be needed.

zmiklank: I see.

zmiklank (approved):

pkubatrh:
Thanks! Merging

PR #396, opened 2023-02-21
remove EOL versions

zmiklank:
This code is no longer maintained.

I've noticed, that for example `test/run` files are copied and not symlinked to the main `test/run` and that makes them un-maintanable.

Do you think we can remove the EOL versions?


zmiklank:
[test-all]

phracek:
[test-openshift]

phracek:
[test-all]

phracek:
[test-openshift]

phracek (changes requested):
It would be nice to have in comment, what versions have been removed.

zmiklank:
> It would be nice to have in comment, what versions have been removed.

Done.

phracek (approved):
LGTM. Thanks for removing deprecated versions.

Issue #400, opened 2023-03-13
Run script error


paulbadcock:
When running an s2i compiled container like below docker file

```dockerfile
FROM registry.access.redhat.com/ubi9/php-81
USER 0
ADD . /tmp/src
RUN chown -R 1001:0 /tmp/src
USER 1001

# Install the dependencies
RUN /usr/libexec/s2i/assemble

# Set the default command for the resulting image
CMD /usr/libexec/s2i/run
```

You will get the following errors
```shell
/usr/libexec/s2i/run: line 34: /etc/php.ini: Permission denied
/usr/libexec/s2i/run: line 35: /etc/php.d/10-opcache.ini: Permission denied
```

If you check inside the container
```shell
sh-5.1$ ls -la /etc/php.ini 
-rw-r--r--. 1 root root 62986 Jul  6  2022 /etc/php.ini

Lines of the the /usr/libexec/s2i/run
```shell
envsubst < /opt/app-root/etc/php.ini.template > ${PHP_SYSCONF_PATH}/php.ini
envsubst < /opt/app-root/etc/php.d/10-opcache.ini.template > ${PHP_SYSCONF_PATH}/php.d/10-opcache.ini
```

Some more info
```
docker run -it registry.access.redhat.com/ubi9/php-81 ls -la /etc/php.ini
-rw-r--r--. 1 root root 62986 Jul  6  2022 /etc/php.ini

docker run -it registry.access.redhat.com/ubi8/php-72 ls -la /etc/php.ini
-rwxrwxrwx. 1 root root 62221 Nov  8  2019 /etc/php.ini

docker run -it registry.access.redhat.com/ubi8/php-74 ls -la /etc/php.ini
-rwxrwxrwx. 1 root root 63599 Jul  7  2022 /etc/php.ini
```

You can resolve it via:
```dockerfile
FROM registry.access.redhat.com/ubi9/php-81
USER 0
ADD . /tmp/src
RUN chown -R 1001:0 /tmp/src
RUN chmod 777 /etc/php.ini /etc/php.d/10-opcache.ini
USER 1001

# Install the dependencies
RUN /usr/libexec/s2i/assemble

# Set the default command for the resulting image
CMD /usr/libexec/s2i/run
```

But something is wrong in those files permissions in the container build.

pkubatrh:
Thanks for the report! Sounds like a side-effect of #386. Will follow up with a PR change the permissions of these specific files.

paulbadcock:
[...]

PR #401, opened 2023-03-14
set correct permissions for config files

pkubatrh:
Fixes: #400


pkubatrh:
[test-all]

phracek (approved):
Everything LGTM. Thanks for the fix. Let's wait for tests and especially OpenShift 4 tests.

pkubatrh:
Had to drop a patch as it was already added in a different PR + rebased on master. Let's hope the tests are still green

[test-all]

Issue #403, opened 2023-03-15
Environment variables are being ignored by the php engine


olstjos:
Our application is crashing because we rely on environment variables for the 
DB_HOST, 
DB_PASSWORD, 
DB_USER
 and other important environment variables for AWS that we cannot push into source control with our app build.

This regression appears to be related to changing from mod_php to php_fpm

I found a Red Hat support article but I have to ask colleagues to open it because I do not have the red hat subscription, someone else does.
https://access.redhat.com/solutions/6972818

Ubi9/php-81 image change from mod_php to php_fpm drops environment variables.

So I found this :

https://github.com/docker-library/php/issues/74

Suggests to add **clear_env = no** to the conf for php-fpm

Ok so I'm attempting to fix it as follows:

```
diff --git a/Dockerfile b/Dockerfile
index 148dbef..a13ca34 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,9 @@ RUN curl -s -o $COMPOSER_FILE https://getcomposer.org/installer && \
 
 COPY / /opt/app-root/src
 
+# Ensure that environment variables are maintained, add it to an existing [www] pool.
+RUN echo "clear_env = no" >> /etc/php-fpm.d/www.conf
+
RUN chmod -R ug=u+wX /etc/php.d && \
     chmod ug=u+wx /etc/php.ini && \
     chmod 777 /var/log && \

```

olstjos:
this fix works, the fix should likely be incorporated into the s2i-php-container

pkubatrh:
This was reported for ubi8/php-80 as well: https://bugzilla.redhat.com/show_bug.cgi?id=2120351

PHP experts say that this change of behavior is made by upstream and is security related and as such I am not sure if removing it should get into the images by default - we tend to follow what upstream thinks is best configuration. Feel free to add a comment to the Bugzilla bug however if you disagree, it's likely there most of the discussion will happen if we see more folks hitting this issue.

olstjos:
We're using environment variables for "security reasons" justified as follows:

Our application needs secrets stored in open shift that we cannot commit into source control.

Our application does not permit unprivileged users from executing random php code.

paulbadcock:
I also really dislike this default. As a compromise can  a pattern to generate the www.conf like the php.ini https://github.com/sclorg/s2i-php-container/blob/master/8.1/s2i/bin/run#L34 ?

[...]

PR #406, opened 2023-04-19
Support clear_env variable by PHP_CLEAR_ENV

phracek:
This pull request fixes issue #403.

By parameter `PHP_CLEAR_ENV=OFF` user can set 'clear_env = false' in /etc/pphp-fpm.d/www.conf file.


phracek:
[test]

phracek:
@pkubatrh @hhorak @remicollet Tests passed. Can you please have a look at it and made a review? Your review is more than appropriate. Thanks. The change is valid only for RHEL9 and only in case of `PHP_CLEAR_ENV` is set to `true`.

hhorak:
It would be good to include more info in the commit messages, a summary from https://github.com/sclorg/s2i-php-container/issues/403 might work to explain basic background why clearing environment is not desired sometimes.

phracek:
> e should likely document this new variable in the README.md, shouldn't we?

README.md files are updated. Thanks for the point.

phracek:
> It would be good to include more info in the commit messages, a summary from #403 might work to explain basic background why clearing environment is not desired sometimes.

I have tried to summarize it somehow. If something is still unclear, please let me know.

phracek:
[test]

phracek:
[test]

phracek:
@remicollet Hi Remi, once you have free or spare time. Can you please have a look at this pull request? Thank you.

8.0/Dockerfile.rhel9
```diff
@@ -60,6 +60,7 @@ ENV PHP_CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/php/ \
     PHP_FPM_CONF_D_PATH=/etc/php-fpm.d \
     PHP_FPM_CONF_FILE=www.conf \
     PHP_FPM_RUN_DIR=/run/php-fpm \
+    PHP_CLEAR_ENV=false \
```
pkubatrh, line 63: We should default to clear_env being `true`, as that is what upstream recommends. The variable should instead be used to override the default setting of `true` to `false`

pkubatrh: I would also recommend to switch the choices to ON/OFF, similar to existing variables.

phracek: > I would also recommend switching the choices to ON/OFF, similar to the existing variables.

Thanks for the review, but I don't follow this comment. If and only if PHP_CLEAR_ENV is defined as 'true' than the variable is set.

phracek:
[test]

8.0/Dockerfile.rhel9
pkubatrh: See eg. https://github.com/sclorg/s2i-php-container/blob/master/8.0/README.md?plain=1#L166
The variable is a boolean and has either `ON` or `OFF` value. We should mimic this behavior so that it is the same across all our boolean variables.
Does it make more sense now?

pkubatrh: Though I guess these variables are using the same values as the configuration files are using :/ Maybe instead make a new section specific to fpm (adn add the new variable there) and use `true` and `false` same as the fpm config is using?

8.0/root/usr/libexec/container-setup
```diff
@@ -67,6 +67,8 @@ if [ "x$PLATFORM" == "xel9" ] || [ "x$PLATFORM" == "xfedora" ]; then
   fi
   if [ "${PHP_CLEAR_ENV:-false}" == "true" ]; then
     echo "clear_env = yes" >> "${PHP_FPM_CONF_D_PATH}/${PHP_FPM_CONF_FILE}"
+  else
+    sed -e 's/^(clear_env)\s+.*/clear_env = no/' -i ${PHP_FPM_CONF_D_PATH}/${PHP_FPM_CONF_FILE}
```
pkubatrh, line 71: Just wondering - why do we explicitly check for existing `clear_env`  definitions here and not in the `true` case?

xrow:
@phracek @remicollet @hhorak I would love to see this merged... I think it got stuck somewhere in the **nirvana**... :-)

phracek:
I am gonna finish this next week, so also tests are passing. Please be patient.

xrow:
Actually I didn`t see that someone wokred on it last week. My bad.

phracek:
[test-openshift]

phracek:
[test-openshift]

phracek:
@remicollet @pkubatrh @hhorak Please review it. So we can merge it.

phracek:
[test]

remicollet:
I don't understand the change

- why making FPM configuration word writable?

- about clear_env

We already have in https://github.com/sclorg/s2i-php-container/blob/master/8.1/root/usr/share/container-scripts/php/common.sh#L47

```
    sed -e 's/^(clear_env)\s+.*/clear_env = no/' -i ${PHP_FPM_CONF_D_PATH}/${PHP_FPM_CONF_FILE}
```

This sed comment probably need to be fixed, as the default line in RPM provided file is

```
;clear_env = no
```

phracek:
> fixed

Thanks for the review, I will fix it. You are right.

phracek:
> We already have in https://github.com/sclorg/s2i-php-container/blob/master/8.1/root/usr/share/container-scripts/php/common.sh#L47
> 
> ```
>     sed -e 's/^(clear_env)\s+.*/clear_env = no/' -i ${PHP_FPM_CONF_D_PATH}/${PHP_FPM_CONF_FILE}
> ```
> 
> This sed comment probably need to be fixed, as the default line in RPM provided file is
> 
> ```
> ;clear_env = no
> ```

@remicollet Good comment. In case of PHP_CLEAR_ENV is set to `false`, than I add to configuration file at the end `clear_env = no`. 
There are three posibbilities.
* `clear_env` is not present in www.conf file and then it should be added
* `;clear_env = no` PHP_CLEAR_ENV is set to yes then clear_env = no is added at the end.
* `clear_env = no` is present and then it will be mentioned twice in configuration file.

Do we want to handle all those states? The condition that will handled all those cases would be a bit complicated. I did not try it yet ;) But can try

remicollet:
AFAIK` ;clear_env = no` is present in configuration seems exists (PHP 5.6)

IIUC the current  code, is that we always expect clear_env to be disabled.

Perhaps a simple fix of the sed is enough.

```
- sed -e 's/^(clear_env)\s+.*/clear_env = no/' -i ${PHP_FPM_CONF_D_PATH}/${PHP_FPM_CONF_FILE}
+ sed -e 's/^[;]*\s*clear_env\s*=.*/clear_env = no/'  -i ${PHP_FPM_CONF_D_PATH}/${PHP_FPM_CONF_FILE}
```

And If you really want a config, `PHP_CLEAR_ENV` being true to disable the feature seems confusing

Also IMHO it should also allow both value

so:

```
if [ "${PHP_CLEAR_ENV:-false}" == "false" ]; then
    sed -e 's/^[;]*\s*clear_env\s*=.*/clear_env = no/'  -i ${PHP_FPM_CONF_D_PATH}/${PHP_FPM_CONF_FILE}
else
    sed -e 's/^[;]*\s*clear_env\s*=.*/clear_env = yes/'  -i ${PHP_FPM_CONF_D_PATH}/${PHP_FPM_CONF_FILE}
fi
```

phracek:
Rebased and added also to 8.2 version.

phracek:
[test]

phracek:
[test-openshift]

phracek:
Tests are passing. Completely.

remicollet (changes requested):

test/run
```diff
@@ -181,6 +197,11 @@ test_application() {
 
   test_config_writeable
   ct_check_testcase_result $?
+
+  if [ "${OS}" == "rhel9" ] && [ "${PHP_CLEAR_ENV:-true}" == "false" ]; then
```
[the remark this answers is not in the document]
phracek: Grrrr, I double check my pull request code and I have missed it.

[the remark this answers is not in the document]
phracek: Good catch.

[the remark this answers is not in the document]
phracek: Fixed by https://github.com/sclorg/s2i-php-container/pull/406/commits/5eabbed6c96b43acff2eb8173b55eb723f9ec3fe

phracek:
[test]

remicollet (approved):
LGTM

xrow:
Hi @phracek,

sorry for my late reply, but I just noticed it. It looks like having this functionality only in run script should be sufficent for all use cases. No need for functionality container-setup (unless it is also prt of some run activity) and assemble, but maybe there are reasons for it. Just a comment. No need to fix.

PR #410, opened 2023-05-11
fix permissions for fpm configuration

pkubatrh:

### DIFF
--- a/8.0/root/usr/libexec/container-setup
+++ b/8.0/root/usr/libexec/container-setup
@@ -51,6 +51,7 @@ if [ "$PLATFORM" == "el7" ]; then
 else
   chmod a+rwx "${PHP_SYSCONF_PATH}/php.ini"
   chmod -R a+rwx "${PHP_SYSCONF_PATH}/php.d"
+  chmod -R a+rwx "${PHP_SYSCONF_PATH}/php-fpm.d"
 fi
 
 if [ "x$PLATFORM" == "xel9" ] || [ "x$PLATFORM" == "xfedora" ]; then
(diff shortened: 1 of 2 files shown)


pkubatrh:
Follow up for #401 as I missed some fpm configuration in 8.0 and 8.1

phracek:
The fix is part of #406. Only for info.

phracek (changes requested):
Everything LGTM. What about adding as well www.conf file like:
`chmod a+rwx ${PHP_FPM_CONF_D_PATH}/${PHP_FPM_CONF_FILE}`
${PHP_SYSCONF_PATH}/php-fpm.d equals to ${PHP_FPM_CONF_D_PATH}.

Can you please modify it?

pkubatrh:
> What about adding as well [www.conf](http://www.conf/) file like: chmod a+rwx ${PHP_FPM_CONF_D_PATH}/${PHP_FPM_CONF_FILE}

The chmod call is recursive, so that should not be needed. I have modified the original call to use `${PHP_FPM_CONF_D_PATH}` however.

pkubatrh:
[test]

phracek (changes requested):
Please address my changes.

8.0/root/usr/libexec/container-setup
```diff
@@ -51,6 +51,7 @@ if [ "$PLATFORM" == "el7" ]; then
 else
   chmod a+rwx "${PHP_SYSCONF_PATH}/php.ini"
   chmod -R a+rwx "${PHP_SYSCONF_PATH}/php.d"
+  chmod -R a+rwx "${PHP_SYSCONF_PATH}/php-fpm.d"
 fi
 
 if [ "x$PLATFORM" == "xel9" ] || [ "x$PLATFORM" == "xfedora" ]; then
```
phracek, line 57:
```suggestion
if [ "x$PLATFORM" == "xel9" ] || [ "x$PLATFORM" == "xfedora" ]; then
chmod -R a+rwx "${PHP_SYSCONF_PATH}/php-fpm.d"
```

phracek:
[test]

phracek (approved):
LGTM. Thanks for the fix

PR #422, opened 2023-08-28
Enable building and pushing image to Quay.io

phracek:


zmiklank (approved):
LGTM, thanks. I also checked if the repository exists in quay.io and it does.

PR #425, opened 2023-10-16
Get rid off obsoleted exclude files

phracek:
Get rid off obsoleted exclude files

Deleted in `7.3` version:

* `.exclude-c9s` - `Dockerfile.c9s` does not exist

Deleted in `7.4` version:

* `.exclude-c9s` - `Dockerfile.c9s` does not exist


phracek:
[test]

zmiklank (approved):
LGTM.

PR #409, opened 2023-05-09
PHP 8.1 for CS9

xrow:
Hi,

I made php 8.1 build for CS9. Please review.

PR #437, opened 2024-02-27
Fixes for Dockerfile.c9s

xrow:
@zmiklank

I fucked up on https://github.com/sclorg/s2i-php-container/pull/409 and closed it by accident.

Can you merge this?


zmiklank (approved):
LGTM, thanks for the fix.

PR #441, opened 2024-03-27
Imagestream generator

zmiklank:
This PR removes tag "7.3" as it not standard for other sclorg images. I also removes tring `"supports":"php:8.0,php"` from the tag. Could that cause a problem?

### DIFF
--- a/.github/workflows/openshift-tests.yml
+++ b/.github/workflows/openshift-tests.yml
@@ -3,10 +3,24 @@ on:
     types:
       - created
 jobs:
+  check-imagestreams:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      statuses: write
+    if: |
+      github.event.issue.pull_request
+      && (contains(github.event.comment.body, '[test-openshift]') || contains(github.event.comment.body, '[test-all]'))
+      && contains(fromJson('["OWNER", "MEMBER"]'), github.event.comment.author_association)
+    steps:
+      - uses: sclorg/ci-scripts/ocp-stream-generator@master
+        with:
+          ref: "refs/pull/${{ github.event.issue.number }}/head"
+
   openshift-tests:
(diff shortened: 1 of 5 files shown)


zmiklank:
[test-openshift]

zmiklank:
[test-openshift]

zmiklank:
@phracek seems like problem with ansible scripts?

phracek:
> @phracek seems like problem with ansible scripts?

No problem with subscription: ```"The system is unable to complete the requested transaction (HTTP error code 500: Internal Server Error)"```

phracek:
[test-openshift]

phracek (approved):
LGTM

zmiklank:
We need an answer for the first question before merging this:

> This PR removes tag "7.3" as it not standard for other sclorg images. I also removes tring "supports":"php:8.0,php" from the tag. Could that cause a problem?

zmiklank:
> We need an answer for the first question before merging this:
> 
> > This PR removes tag "7.3" as it not standard for other sclorg images. I also removes tring "supports":"php:8.0,php" from the tag. Could that cause a problem?

@pkubat WDYT?

phracek:
PHP-7.3 is still supported for RHEL7 images. I don't follow what you mean by removing tag '7.3'?

zmiklank:
> PHP-7.3 is still supported for RHEL7 images. I don't follow what you mean by removing tag '7.3'?

There is "7.3-ubi7" and a "7.3" tag. Those are two different tags we provide, the "7.3" being non-standard - not all container images' imagrestreams provide it. Currently the imagestream generator can not generate tags in format of app version only (7.3).

pkubatrh:
LGTM. The change might cause some minor breakege, but since we are standardizing on a specific name format I say this is worth the risk.

phracek (approved):
LGTM again.

PR #444, opened 2024-05-21
re-add previously removed "7.3" IS tag

zmiklank:
we need to maintain it for the lifetime of this image.


zmiklank:
[test-openshift]

phracek (changes requested):
Aarch64 is missing, It is still supported as well.
https://access.redhat.com/support/policy/updates/rhscl-rhel7

zmiklank:
We do not provide imagestreams for RHEL 7 for aarch64 at all.

PR #454, opened 2024-07-08
Move $HOME env variable outside web server's Document root

### LINKED ISSUE #255: HOME directory
Hello. I am using centos/php-72-centos7 as a base template for my Dockerfile.

When I run an interactive terminal into the container, it creates .bash_history file inside APP_DATA directory which I don't want.

I have tried changing the home directory in /etc/passwd but this does not seem to make any difference. When I run `echo $HOME`, it is always set to what $APP_DATA is.

I know I can override the HOME variable by setting the environment variable to the docker exec command, but I'd prefer not having to do this specifically for this image container.

Is there a way to change the HOME directory for the image during the build process or when it is run by docker-compose?

SlouchyButton:
Overwrite $HOME env variable for all supported versions. Default $HOME is directed inside the web server's Document root, which means that applications that save potentially private data to $HOME (e.g. bash's histfile) will save them into a folder accessible outside the container via the web server. This means there is a possibility of leaking the data.

This does not occur in all cases, namely bash won't create a histfile at all, when user sets a different user via `--user=` argument in `podman run` command, as in that case bash doesn't have a permission to write into the $HOME.

Fixes: #255

### DIFF
--- a/test/run
+++ b/test/run
@@ -171,7 +171,7 @@ test_home_not_in_webroot() {
   test_connection ${test_port}
 
   info "Testing whether \$HOME points to web server Document root"
-  homef=$(docker exec -it $(cat ${cid_file}) /bin/bash -c "echo \$HOME")
+  local homef=$(docker exec -it $(cat ${cid_file}) /bin/bash -c "echo \$HOME")
   ct_check_testcase_result $?
   if [ "${homef}" = "${homef##/opt/app-root/src}" ]; then
     echo "Home folder not set to Document root"
--- a/test/run
+++ b/test/run
@@ -19,6 +19,7 @@ test_ssl
 test_ssl_own_cert
 ct_npm_works
 test_build_from_dockerfile
+test_home_not_in_webroot
 "
 
 TEST_CLEAR_ENV="\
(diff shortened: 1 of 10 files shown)


pkubatrh:
[test]

test/run
```diff
@@ -162,6 +163,25 @@ test_config_writeable() {
...
 
+test_home_not_in_webroot() {
+  cid_file=$CID_FILE_DIR/$(mktemp -u -p . --suffix .cid)
+  docker run -d --cidfile=${cid_file} ${IMAGE_NAME}-testapp
+
+  # Wait for container to initialize fully
+  test_connection ${test_port}
+
+  info "Testing whether \$HOME points to web server Document root"
+  homef=$(docker exec -it $(cat ${cid_file}) /bin/bash -c "echo \$HOME")
```
phracek, line 174: please use at the beginning `local homef` so that variable is not expanded into whole script.

Otherwise LGTM.

phracek (changes requested):

test/run
SlouchyButton: Changed

phracek:
[test-all]

phracek (commented):
Thanks for this pull request. The change itself LGTM. Let's wait for the tests. Good Job.

phracek:
I don't know what is the status here. Let's convert it to Draft.

PR #456, opened 2024-07-17
Fix calling proper python interpreter

phracek:
This pull request fixes calling proper python interpreter.


phracek:
[test-openshift-pytest]

PR #474, opened 2025-01-15
Enable testing s2i-php-container on RHEL10 host

phracek:
No code is changed by this pull request.


phracek:
Let's test it only for sure

[test]

PR #476, opened 2025-01-28
Deprecation PHP-80 image on RHEL8

phracek:
This pull request deprecates PHP-80 image from RHEL8.

It reached EOL in Nov 2024.

- The first commit adds `.exclude-rhel8` file into 8.0 directory
- The documentation is updated with deprecation 7.3 which is completely EOL. It is removed also from Makefile.

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


phracek:
[test]

phracek:
@remicollet PTAL. Thanks. Approval is only needed.

remicollet (approved):
LGTM

PR #479, opened 2025-01-30
use ext name instead of pkg name, and add redis

remicollet:
Main benefit: will install the extension, whatever the package name is, and latest version, ex:

- zip => php-zip => php-pecl-zip
- json => php-pecl-json => php-json => php-common (since 8.0)
- redis => php-pecl-redis3 => php-pecl-redis5 => php-pecl-redis6
- etc

By packaging Guidelines, the `php-<extname>` is always provided and is the only reliable name
https://docs.fedoraproject.org/en-US/packaging-guidelines/PHP/#requires-provides-c

### DIFF
--- a/8.3/Dockerfile.c10s
+++ b/8.3/Dockerfile.c10s
@@ -44,7 +44,7 @@ LABEL summary="${SUMMARY}" \
 ARG INSTALL_PKGS="php php-fpm php-mysqlnd php-pgsql php-bcmath \
                   php-gd php-intl php-ldap php-mbstring php-pdo \
                   php-process php-soap php-opcache php-xml \
-                  php-gmp php-pecl-apcu php-pecl-zip mod_ssl hostname"
+                  php-gmp php-apcu php-zip php-redis mod_ssl hostname"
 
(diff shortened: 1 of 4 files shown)


phracek:
[test]

phracek:
@remicollet IT looks like on CentOS Stream 10 packages do not exist

```
package php-apcu is not installed
package php-zip is not installed
package php-redis is not installed
```

remicollet:
[test]

remicollet:
> @remicollet IT looks like on CentOS Stream 10 packages do not exist
> 
> ```
> package php-apcu is not installed
> package php-zip is not installed
> package php-redis is not installed
> ```

Yes, because these are NOT package name (so rpm -V fails)
Second commit try to fix this

remicollet:
```
Found VERSION 8.3
Found requested extensions

```
:)

phracek:
Thanks. Let's try another shot

[test]

phracek:
@remicollet One more question. The Dockerfile.rhel10 is not valid? Thanks for the great work, though.

phracek (approved):
Thanks for fixing. LGTM.

PR #482, opened 2025-04-14
s2i-core and s2i-base images are available from unauthenticated registries

phracek:
Certificates are not need at all in Dockerfile.
Remove these certificates.

See JIRA: https://issues.redhat.com/browse/ENGCMP-5283


phracek:
[test]

PR #486, opened 2025-05-15
Add new imagestreams for RHEL10 and RHEL9.6.

phracek:
Add new imagestreams for RHEL10 and RHEL9.6.

Update all OpenShift 4 and OpenShift 4 Pytest test suites.
Test `test_curl_connection` does not make sense for Helm Chart. So remove it.


phracek:
Everything is done and GREEN. No code is changed. Let's get merge it.

PR #489, opened 2025-06-12
Fix regular expression for proper return 'rhel' name from 'rhel10'

phracek:
Fix regular expression for proper return 'rhel' name from 'rhel10'

All expression returns 'rhel1'. The new expression returns proper name


phracek:
[test-openshift]

phracek:
[test-openshift-pytest]

PR #491, opened 2025-07-02
Do not test PHP Openshift tests

phracek:
It reached EOL in May 2025.


phracek:
[test-openshift][test-openshift-pytest]

phracek:
Let's test again. Helm chart was renamed
[test-openshift][test-openshift-pytest]

phracek:
This code does not break any PHP container fucntionality. Only fixes tests.

PR #492, opened 2025-07-03
Remove PHP-81 from GitHub Actions. It is not supported at all

phracek:


phracek:
Running tests is not needed.

